PR #57, opened 2026-05-22
fix(games): resolve quick math challenge loopholes

### LINKED ISSUE #46: Bug: Quick Math Challenge: Score Exploit via Toast & Zero Input Hidden
**Type:** Gameplay Loophole · Input UX Bug
**Area:** Brain Games — Quick Math Challenge
**File:** `BrainGames.tsx`

---

### Description

The Quick Math Challenge has two bugs in the answer submission flow — one that allows trivial score inflation and one that silently hides a valid answer from the user.

---

### Bug 1 — Score Exploit via Answer Toast

When an incorrect answer is submitted, `generateMathQuestion()` is not called. The user stays on the same question while the error toast reveals the correct answer. The user can immediately re-submit that answer for a point.

```ts
// ❌ BrainGames.tsx — question not refreshed on wrong answer
} else {
  showError("✗ Incorrect", `The answer was ${correct}. Keep practicing!`);
  // generateMathQuestion() is missing — same question persists
}
```

**Steps to reproduce:**
1. Open Quick Math Challenge.
2. Submit any incorrect answer.
3. Read the correct answer from the error toast.
4. Re-submit it immediately for a free point.

---

### Bug 2 — Input Hides `0` Due to Falsy `||` Binding

The answer input is bound using:

```ts
// ❌ BrainGames.tsx — falsy evaluation swallows the digit 0
value={mathQuestion.answer || ""}
```

When a user types `0`, the expression `0 || ""` evaluates to `""` because `0` is falsy in JavaScript. The input appears blank — the digit is invisible to the user, causing confusion and failed submissions.

---

### Impact

- Any user can trivially inflate their Quick Math score by reading the correct answer from the error toast and re-submitting it.
- Users whose correct answer is `0` see an empty input field and cannot confirm their entry, leading to incorrect re-submissions and a frustrating UX.

---

### Proposed Fix

**Bug 1:** Call `generateMathQuestion()` immediately after the error toast so the question always advances on an incorrect attempt:

```ts
// Advance to a new question on wrong answer
} else {
  showError("✗ Incorrect", `The answer was ${correct}. Keep practicing!`);
  generateMathQuestion();
}
```

**Bug 2:** Store user input in a dedicated `string` state and use `??` (nullish coalescing) instead of `||` so `0` is preserved:

```ts
// Separate input state — 0 is visible and handled correctly
const [inputValue, setInputValue] = useState<string>("");

<Input
  value={inputValue}
  onChange={(e) => setInputValue(e.target.value)}
/>

// On submit, compare numerically
const userAnswer = Number(inputValue);
```

Using `??` instead of `||` also works if a single state is preferred — `??` only falls back for `null` and `undefined`, not `0`.

---

### Expected Behavior After Fix

- An incorrect answer always advances to a new question — the revealed answer is never reusable.
- Typing `0` into the input field displays and submits correctly.
- Score integrity is maintained across all Quick Math sessions.

---

### Additional Recommendations

- Consider whether the correct answer needs to appear in the toast at all. Even with `generateMathQuestion()` added, a fast user can read the answer before the question changes. Replacing it with a generic "Try again!" message — or showing the answer only after a short delay — closes the loophole more completely.
- Add a submit cooldown or debounce on the answer button to prevent rapid re-submission between question transitions.

Tanish-Solanki:
## Changes

* Fixed gameplay loophole where users could reuse answers revealed in incorrect-answer toasts
* Added automatic question regeneration after incorrect submissions
* Fixed input UX issue where `0` values appeared blank in the answer field
* Preserved existing input behavior while correctly handling falsy numeric values

## Tested

* Verified incorrect answers now generate a new question immediately
* Verified `0` displays correctly in the input field
* Tested answer validation and score flow locally

Closes #46

### DIFF
--- a/src/pages/BrainGames.tsx
+++ b/src/pages/BrainGames.tsx
@@ -20,7 +20,7 @@ const BrainGames = () => {
   const [memoryCards, setMemoryCards] = useState<number[]>([]);
   const [flippedCards, setFlippedCards] = useState<number[]>([]);
   const [matchedCards, setMatchedCards] = useState<number[]>([]);
-  const [mathQuestion, setMathQuestion] = useState({ num1: 0, num2: 0, answer: 0 });
+  const [mathQuestion, setMathQuestion] = useState({ num1: 0, num2: 0, answer: "" });
   const [mathScore, setMathScore] = useState(0);
   const [wordSequence, setWordSequence] = useState<string[]>([]);
   const [userSequence, setUserSequence] = useState<{ word: string; index: number }[]>([]);
@@ -307,7 +307,7 @@ const BrainGames = () => {
   const generateMathQuestion = () => {
     const num1 = Math.floor(Math.random() * 50) + 10;
     const num2 = Math.floor(Math.random() * 50) + 10;
-    setMathQuestion({ num1, num2, answer: 0 });
+    setMathQuestion({ num1, num2, answer: "" });
   };
 
   const handleCardClick = (index: number) => {
@@ -344,7 +344,8 @@ const BrainGames = () => {
 
   const checkMathAnswer = () => {
     const correct = mathQuestion.num1 + mathQuestion.num2;
-    if (mathQuestion.answer === correct) {
+    const userAnswer = parseInt(mathQuestion.answer);
+    if (userAnswer === correct) {
       const newScore = mathScore + 1;
       setMathScore(newScore);
       showSuccess("✓ Correct! ✓", `Score: ${newScore}`);
@@ -360,6 +361,7 @@ const BrainGames = () => {
         description: `The answer was ${correct}`,
         variant: "destructive",
       });
+      generateMathQuestion();
     }
   };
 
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
@Tanish-Solanki Could you please share before and after results of the demonstration?

Tanish-Solanki:
Sure @mohdmaazgani ,
### Before Fix

1. Enter an incorrect answer (example: `5` for `2 + 2`)
2. Toast reveals: `The answer was 4`
3. Same question remains on screen
4. Re-enter `4`
5. Score increments incorrectly without solving a new question

Input issue:

* Typing `0` caused the input field to appear blank because the value binding used `answer || ""`
image


### After Fix

1. Enter an incorrect answer
2. Toast still appears
3. A new math question is generated immediately
4. Previous answer can no longer be reused to gain points

Input issue resolved:

* Typing `0` now displays correctly in the input field
* Answer validation works normally for zero-valued inputs

image
After changes, answer was wrong but still question changed.

### Verification

* Tested multiple incorrect-answer flows locally
* Verified score cannot be artificially inflated
* Verified `0` renders and submits correctly
* Confirmed no regressions in normal gameplay flow

Issue #72, opened 2026-05-23
Problem of layering on the main page


ishitaajain22-tech:
When on the main page, that is the url of the page, once my crediantials is stored in the database, and i click on sign-in there is immediate change and page fluttering and then the dashboard comes up

Fix: The fix needed in both files is , every onClick={() => navigate("/auth")} and every onClick={() => { showInfo(...); navigate("/auth"); }} needs a brief loading state before navigating

Two files to replace: 
src/pages/Index.tsx
src/components/Hero.tsx

PR #80, opened 2026-05-23
Optimize auth state listener in ProtectedRoute

ishitaajain22-tech:
Refactor authentication state handling to prevent flickering.
It prevents the fluttering of the page , as i have adjusted and updated the states accordingly


ishitaajain22-tech:
fixes issue #72

Issue #94, opened 2026-05-24
Bug: Repetition of facts in health Facts


ishitaajain22-tech:
There is quite a repetition of halth facts again and again, if the user is serious and reading it, they will recognize the pattern

<img width="1573" height="857" alt="Image" src="https://github.com/user-attachments/assets/cce209e8-0355-4cb7-a7d4-08c9addcbf45" />

PR #96, opened 2026-05-25
HealthFacts component with new features

ishitaajain22-tech:
fixes issue #94 
1. 35 curated health topics — Cardiovascular disease, Dopamine, CRISPR, Telomere, Gut microbiota, Melatonin, Placebo, Microplastics, and more. Each maps to a real Wikipedia article.

2. No-repeat logic — same Set approach as before. Topics cycle through all 35 before any repeats, then reset.
Wikipedia thumbnail — if the article has an image, it shows as the card thumbnail instead of the icon. Articles like DNA, Cancer, Neuron usually have one.

3."Read full article" link — every fact has a direct link to the Wikipedia page so curious users can go deeper.
Skeleton loading + error state — same as before, fully handled.

### DIFF
--- a/src/pages/HealthFacts.tsx
+++ b/src/pages/HealthFacts.tsx
@@ -1,147 +1,156 @@
 import { useState, useEffect } from "react";
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
-import { Sparkles, RefreshCw, BookOpen, Heart, Brain, Dna } from "lucide-react";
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
@ishitaajain22-tech The health topics are good, but in the next PR, add the "health facts" that sound interesting. These are more of a textbook kind of information.

Include facts that engage people, not big paragraphs that people get bored of. Some interesting and out-of-the-box facts that exist and people are unaware of.

mohdmaazgani:
@ishitaajain22-tech I am merging this PR , and i want to you raise another PR and include those facts as well so that this become an exctiting corner for the readers to read out.

ishitaajain22-tech:
@mohdmaazgani hello
ok i will surely open another

PR #112, opened 2026-05-25
Fix dashboard logo redirect behavior

### LINKED ISSUE #109: Clicking on "Health Tracker" redirects to landing page instead of dashboard
## Description

When clicking on the "Health Tracker" text/logo from the dashboard sidebar, the user gets redirected to the main landing page.

Since the user is already logged in and inside the dashboard, redirecting to the landing page feels slightly confusing and breaks the dashboard flow.

## Current Behavior

* Clicking "Health Tracker" opens the public landing page

## Expected Behavior

* Clicking "Health Tracker" should redirect to the dashboard home page or keep the user inside the dashboard

Image

the-rahul-07:
### Description

This PR fixes the dashboard sidebar logo navigation behavior.

Previously, clicking on the "Health Tracker" text/logo in the sidebar redirected authenticated users to the public landing page, which interrupted the dashboard flow.

### Changes Made
Updated the sidebar logo navigation to redirect users to the dashboard
Ensured authenticated users remain inside the dashboard experience

### Testing
Verified the application runs successfully locally
Tested logo click behavior from dashboard
Confirmed users are redirected to /dashboard instead of the landing page

### Fixes #109 

@mohdmaazgani Could you please assign the appropriate labels to this PR? Thank you!

### DIFF
--- a/src/components/AppSidebar.tsx
+++ b/src/components/AppSidebar.tsx
@@ -53,10 +53,10 @@ export function AppSidebar() {
   return (
     <Sidebar collapsible="icon">
       <div className="flex items-center justify-between p-4 border-b border-sidebar-border">
-        <NavLink to="/" className="flex items-center">
+        <NavLink to="/dashboard" className="flex items-center">
           {!isCollapsed && (
             <h2 className="text-lg font-semibold text-sidebar-foreground cursor-pointer">
-              Health Tracker
+              Health Tracker 
             </h2>
           )}
         </NavLink>
--- a/src/components/AppSidebar.tsx
+++ b/src/components/AppSidebar.tsx
@@ -56,7 +56,7 @@ export function AppSidebar() {
         <NavLink to="/dashboard" className="flex items-center">
           {!isCollapsed && (
             <h2 className="text-lg font-semibold text-sidebar-foreground cursor-pointer">
-              Health Tracker 
+              Health Tracker
             </h2>
           )}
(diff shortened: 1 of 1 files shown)


the-rahul-07:
Hi, @mohdmaazgani

Kindly review this PR whenever possible. Please let me know if any modifications or improvements are required from my side.

Thank you!

mohdmaazgani:
@the-rahul-07 There is already a dashboard option present in that layout, so for now this additional redirection is not really required.

- What I can do is merge this PR, and then you can raise the next issue to redirect it to the homepage (as it is currently now), since that is the current workflow of the website. As you already raised the PR and got it assigned, I don’t want your effort to go to waste.
- The other option is that I close this PR for now.
- The call is yours. 👍

the-rahul-07:
Hi, @mohdmaazgani

Thank you for the clarification regarding the current workflow.

In that case, I think it would be better to close this PR for now to keep the existing navigation behavior consistent with the intended design.

Thank you for the guidance and consideration! 👍

mohdmaazgani:
> Hi, @mohdmaazgani
> 
> Thank you for the clarification regarding the current workflow.
> 
> In that case, I think it would be better to close this PR for now to keep the existing navigation behavior consistent with the intended design.
> 
> Thank you for the guidance and consideration! 👍

**Thank you @the-rahul-07 for your cooperation.**

PR #135, opened 2026-05-27
fix: prevent duplicate Health Facts initialization in Strict Mode

### LINKED ISSUE #115: UX & Performance: Double API call and duplicate toast in HealthFacts.tsx due to React 18 Strict Mode
<html>
<body>
<h2>Summary</h2>
<p>In React 18, Strict Mode intentionally mounts and unmounts components twice in development to surface side-effect bugs. The <code>useEffect</code> in <code>HealthFacts.tsx</code> has no cleanup function, so both mount cycles fire <code>fetchFact()</code> and <code>showInfo()</code> concurrently. This produces two parallel Wikipedia API requests, a race condition on the loading state, and a duplicate "Welcome to Health Facts!" toast notification on every page load.</p>
<hr>
<h2>Status</h2>
<ul>
<li><strong>Severity:</strong> 🟡 Low / Medium — UX noise + unnecessary API consumption</li>
<li><strong>Type:</strong> Race condition / missing effect cleanup</li>
<li><strong>File affected:</strong> <code>src/components/HealthFacts.tsx</code></li>
<li><strong>Related:</strong> None — first report</li>
</ul>
<hr>
<h2>Root Cause</h2>
<h3>The effect as written (no cleanup)</h3>
<pre><code class="language-ts">// HealthFacts.tsx — current (broken)
useEffect(() =&gt; {
  fetchFact(true);
  showInfo(
    "Welcome to Health Facts!",
    "Powered by Wikipedia — real medical knowledge"
  );
}, []);
</code></pre>
<p>React 18 Strict Mode runs this sequence on every mount in development:</p>
<ol>
<li>Component mounts → effect runs → <code>fetchFact()</code> call A starts, toast A fires</li>
<li>React immediately unmounts the component (Strict Mode cleanup simulation)</li>
<li>Component remounts → effect runs again → <code>fetchFact()</code> call B starts, toast B fires</li>
</ol>
<p>Because there is no cleanup function returning from the effect, call A is never cancelled when the component unmounts between cycles. Both requests are in-flight simultaneously and both resolve independently, each attempting to update state on the same component.</p>
<hr>
<h2>Impact</h2>

Symptom | Cause
-- | --
Duplicate "Welcome to Health Facts!" toast on every page load | showInfo() called twice with no guard
Two parallel Wikipedia API requests fired per mount | fetchFact() called twice with no cancellation
Flickering or inconsistent loading state | Both calls race to set isLoading and the fact state
Unnecessary API rate limit consumption | Every page visit costs 2 Wikipedia requests instead of 1
Potential stale state update on unmounted component | Call A may resolve after the component has re-mounted from call B's cycle

<blockquote>
<p>While Strict Mode double-invocation only occurs in development, the absence of a cleanup function is a real bug. Any future trigger that causes the component to unmount mid-fetch (navigation, tab switching, conditional rendering) will produce the same race condition in production.</p>
</blockquote>
<hr>
<h2>Steps to Reproduce</h2>
<ol>
<li>Run the app locally with React 18 and Strict Mode enabled (<code>&lt;React.StrictMode&gt;</code> in <code>main.tsx</code> or <code>index.tsx</code>).</li>
<li>Navigate to the Health Facts page.</li>
<li>Observe the toast notification — it appears <strong>twice</strong> in quick succession.</li>
<li>Open the Network tab in DevTools — two Wikipedia API requests are dispatched simultaneously on mount.</li>
<li>Observe the loading indicator — it may flicker or resolve inconsistently depending on which request wins the race.</li>
</ol>
<hr>
<h2>Expected Behaviour</h2>
<ul>
<li>Exactly <strong>one</strong> Wikipedia API request is made on mount.</li>
<li>Exactly <strong>one</strong> "Welcome to Health Facts!" toast appears per page visit.</li>
<li>Loading state transitions cleanly: idle → loading → resolved.</li>
<li>If the component unmounts before the fetch completes (navigation, conditional render), the in-flight request is cancelled and no state update is attempted.</li>
</ul>
<hr>
<h2>Proposed Fix — <code>HealthFacts.tsx</code></h2>
<h3>Option A — <code>AbortController</code> (recommended)</h3>
<p>Cancels the in-flight request on cleanup. The correct solution for any <code>fetch</code>-based effect. Prevents stale updates in both development (Strict Mode) and production (mid-navigation unmount).</p>
<pre><code class="language-ts">useEffect(() =&gt; {
  const controller = new AbortController();

  fetchFact(true, controller.signal);
  showInfo(
    "Welcome to Health Facts!",
    "Powered by Wikipedia — real medical knowledge"
  );

  return () =&gt; {
    controller.abort(); // cancels the pending fetch if component unmounts
  };
}, []);
</code></pre>
<p><code>fetchFact</code> must be updated to accept and forward the signal to its <code>fetch</code> call:</p>
<pre><code class="language-ts">async function fetchFact(initial: boolean, signal?: AbortSignal) {
  try {
    const res = await fetch(WIKIPEDIA_API_URL, { signal });
    // ... existing handling
  } catch (err) {
    if ((err as Error).name === "AbortError") return; // silently ignore cancellation
    // ... existing error handling
  }
}
</code></pre>
<hr>
<h3>Option B — <code>useRef</code> guard flag (simpler, no <code>fetchFact</code> signature change)</h3>
<p>Prevents the second invocation from doing any work. Simpler if changing <code>fetchFact</code>'s signature is not desirable, but does <strong>not</strong> cancel in-flight network requests on production unmounts.</p>
<pre><code class="language-ts">const hasFetched = useRef(false);

useEffect(() =&gt; {
  if (hasFetched.current) return;
  hasFetched.current = true;

  fetchFact(true);
  showInfo(
    "Welcome to Health Facts!",
    "Powered by Wikipedia — real medical knowledge"
  );
}, []);
</code></pre>
<blockquote>
<p>Prefer Option A for production correctness. Use Option B only if <code>fetchFact</code> cannot be modified in this iteration.</p>
</blockquote>
<hr>
<h2>Acceptance Criteria</h2>
<ul>
<li>[ ] Exactly one Wikipedia API request is made when the Health Facts page mounts — verified in the Network tab</li>
<li>[ ] Exactly one "Welcome to Health Facts!" toast appears per page visit</li>
<li>[ ] Loading state transitions cleanly with no flicker</li>
<li>[ ] Navigating away mid-fetch does not trigger a state update on the unmounted component (no React "Can't perform a state update on an unmounted component" warning)</li>
<li>[ ] Behaviour is consistent between development (Strict Mode) and production builds</li>
<li>[ ] No regression to fact refresh, error handling, or any other <code>fetchFact</code> call sites</li>
</ul>
<hr>
<h2>Notes</h2>
<ul>
<li>This pattern should be audited across other <code>useEffect</code> hooks in the codebase that fire API calls on mount — any without a cleanup function returning an <code>AbortController().abort()</code> carry the same risk.</li>
<li>The Strict Mode double-invoke behaviour is intentional by the React team and will not be disabled. The fix must be in the effect, not in removing <code>&lt;React.StrictMode&gt;</code>.</li>
<li>In production, Strict Mode double-invoke does not occur, but the missing cleanup still causes the race condition any time the component unmounts before a fetch resolves (e.g. fast navigation, React Suspense, conditional rendering).</li>
</ul></body></html>
</body>
</html>

Tanish-Solanki:
## Summary

Fixed duplicate initialization behaviour in `HealthFacts.tsx` caused by React 18 Strict Mode double-invoking `useEffect` during development.

The current implementation uses locally stored health facts instead of API-based fetching, so the issue manifested as duplicate toast notifications and repeated initial fact generation rather than duplicate network requests.

## Changes Made

* Added a `useRef` guard (`hasInitialized`) to prevent duplicate effect execution
* Prevented duplicate `showInfo()` toast notifications on initial mount
* Prevented duplicate initial `getRandomFact()` calls
* Removed invalid `AbortController` / `fetchFact` logic that referenced non-existent API code
* Kept existing fact generation, history tracking, and UI behaviour unchanged

## Testing

* Verified only one welcome toast appears on page load
* Verified only one initial health fact is generated
* Confirmed no duplicate history entries are added during initialization
* Tested behaviour under React 18 Strict Mode
* Confirmed no regression to “Show Me Another Fact” functionality

## Screenshots

image

image

image

closes #115


mohdmaazgani:
@Tanish-Solanki Please resolve the merge conflicts and share clear before-and-after screenshots of the changes. A similar implementation was already worked on previously, so the screenshots will help in comparing both approaches and make the review/merge decision easier.
Also, make sure your branch is updated with the latest commits from the main branch so you can incorporate the recent changes made by the other contributor as well.

Tanish-Solanki:
Hi @mohdmaazgani,
Resolved the merge conflicts and updated the branch with the latest changes from `main`.
I also verified the Strict Mode fix after merging the latest updates. The attached screenshot shows the corrected behaviour where the Health Facts page initializes only once, displays a single welcome toast, and continues to function normally without duplicate initialization.

Screenshot 2026-05-28 081629

mohdmaazgani:
@Tanish-Solanki I have assigned 3 different `type` labels to this PR. Could you please clarify whether this improves your contribution scoring, or does the scoring remain the same?

Are points awarded based on:

* all assigned labels combined, or
* only the highest-level type label (for example: `feature`, `refactor`, etc.)?

Would appreciate the clarification.

Tanish-Solanki:
Hi @mohdmaazgani,

Thanks! I was curious because I noticed the dashboard seems to count multiple type labels separately for the type bonus section.

For example, PRs with type:refactor, type:feature, and type:design together appear to receive a combined bonus instead of only the highest-level type label being considered.

image

mohdmaazgani:
@Tanish-Solanki Thank you so much for the updates.

PR #201, opened 2026-05-31
fix: resolve off-by-one error and race condition in rate limiter

### LINKED ISSUE #200: fix: resolve off-by-one error and race condition in rate limiter
# Description

The rate limiter in the symptom-analyzer Deno function contains an off-by-one error and a race condition that allows requests to bypass the rate limit. This can be exploited for DoS attacks against the expensive AI API endpoint.

## Expected Behaviour

- Exactly 10 requests should be allowed per IP per 60-second window
- The 11th request should be blocked
- If Redis is unavailable, the in-memory fallback should maintain the same rate limit

## Actual Behaviour

- 11 requests are allowed per IP per 60s window (off-by-one error)

## Steps to Reproduce

1. Start the Deno function with `UPSTASH_REDIS_REST_URL` and `UPSTASH_REDIS_REST_TOKEN` configured
2. Send 11 rapid requests from the same IP address
3. Observe that all 11 requests succeed (_should only allow 10_)

## Environment

- Deno runtime, Supabase + Upstash Redis
- Affected Component: `symptom-analyzer`

xrvnd:
Closes #200 

## What & Why

Two bugs in `supabase/functions/symptom-analyzer/rateLimit.ts`:

- **Race condition:** `INCR` and `EXPIRE` were separate round-trips. A crash between them left the Redis key with no TTL, permanently rate-limiting that IP.
- **Off-by-one:** `count === 1` was a fragile proxy for "first request." Any edge case that skipped it would silently drop the TTL.

## Fix

Replace both calls with an atomic Redis pipeline using `INCR` + `EXPIRE NX`. `EXPIRE NX` sets the TTL once (only if not already set), removing the `count === 1` dependency entirely. Also unified two duplicate branches in `memoryRateLimit` into a single guard.

### DIFF
--- a/supabase/functions/symptom-analyzer/rateLimit.ts
+++ b/supabase/functions/symptom-analyzer/rateLimit.ts
@@ -27,35 +27,18 @@ if (redisUrl && redisToken) {
 
 function memoryRateLimit(ip: string): { success: boolean } {
     const now = Date.now();
-
     const existing = requestStore.get(ip);
 
-    // First request
-    if (!existing) {
-        requestStore.set(ip, {
-            count: 1,
-            timestamp: now,
-        });
-
+    // Initialize or reset expired window
+    if (!existing || now - existing.timestamp > WINDOW_SIZE_MS) {
+        requestStore.set(ip, { count: 1, timestamp: now });
         return { success: true };
     }
 
-    // Reset window
-    if (now - existing.timestamp > WINDOW_SIZE_MS) {
-        requestStore.set(ip, {
-            count: 1,
-            timestamp: now,
-        });
-
-        return { success: true };
-    }
-
-    // Block request
     if (existing.count >= MAX_REQUESTS) {
         return { success: false };
     }
 
-    // Increment safely
     requestStore.set(ip, {
         count: existing.count + 1,
         timestamp: existing.timestamp,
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
Thanks for the improvement! The Redis pipeline approach and the TTL handling look good.
I just have one clarification before merging. Could you confirm which Redis client/library this function is using and what the exact return shape of `pipeline.exec()` is in that client?

I noticed we're doing:

```ts
const results = await pipeline.exec();
const count = results[0] as number;
```

and I want to make sure `results[0]` is guaranteed to be the incremented count across the Redis client we're using, since different clients can return pipeline results in slightly different formats.

If you've tested this against the project's current Redis setup, could you share the expected response structure or add a small note/test case for it?

xrvnd:
thanks for the review :)

the library is `@upstash/redis@1.38.0` loaded through `esm.sh` . Its `pipeline.exec()` returns a flat array of deserialized results in command order (not `[error, result]` tuples like `ioredis` . 

`results[0]` is directly the `number` returned by INCR : the cast was safe. I will need to push a follow up fix commit that makes this explicit by using typed generic overload --> `const [count] = await pipeline.exec<[number, number]>();`

this removes the `as number` cast entirely and documents the expected shape inline.

mohdmaazgani:
Thanks for the clarification.
That makes sense. My main concern was the return shape of pipeline.exec() since it differs across Redis clients, but if this is using @upstash/redis and you've verified that exec() returns a flat array of deserialised results, then results[0] is indeed safe.
I agree that using the typed overload and destructuring is cleaner and makes the expected response shape much more explicit.
Thanks for the detailed explanation.

mohdmaazgani:
Thank you for contributing to the project! 🌟
If you find the repository helpful, please consider giving it a star ⭐ — it really helps support and motivate the project.
Happy Contributing 🚀

PR #339, opened 2026-06-13
fix: add Contact Support link in Privacy Policy Section 7

### LINKED ISSUE #230: Privacy Policy: Replace Public GitHub Issues with a Private Contact Method for Data Requests
## Description

The Privacy Policy currently directs users to submit privacy-related questions and data requests through the project's public GitHub Issues page.

```tsx
<a
  href="https://github.com/mohdmaazgani/symptom-scribe-clean/issues"
  target="_blank"
  rel="noopener noreferrer"
>
  Open a GitHub Issue
</a>
```

### Problem

Privacy-related inquiries may contain sensitive personal information, including account details and health-related data. Since GitHub Issues are public by default, users may unintentionally disclose sensitive information when exercising their privacy rights.

### Expected Behavior

Users should be provided with a private communication channel for privacy and data-protection requests, such as:

* A dedicated support email address
* A contact form
* A private help desk or ticketing system

### Why This Matters

* Protects user privacy and sensitive information
* Reduces the risk of accidental public disclosure
* Aligns better with common privacy-policy practices
* Improves user trust when submitting data-related requests

### Suggested Solution

Replace the GitHub Issues link in the "Contact Us" section with a private contact method or add a clear warning that GitHub Issues are publicly visible.

### Acceptance Criteria

* [ ] Privacy-related contact information uses a private communication channel
* [ ] Users are not directed to submit sensitive information through public GitHub Issues
* [ ] Privacy Policy is updated accordingly

Tharsiga-21:
closes #230

## Changes Made
- Added a "Contact Support" hyperlink alongside the existing "Open a GitHub Issue" link in Section 7 (Contact Us) of the Privacy Policy page.

## Why
Users previously had only one way to reach out (GitHub Issues). This change provides an alternative contact method through the website's own support page.

## What Changed
- File: `src/pages/Privacy.tsx`
- Section 7 text updated from:
  `please contact us at Open a GitHub Issue`
  to:
  `please contact us at Open a GitHub Issue or Contact Support`
- "Contact Support" links to `/contact` (internal page)
- Matches existing link styling (`text-primary hover:underline`)


mohdmaazgani:
@Tharsiga-21 The changes are well implemented, but it is not working in dark mode. Please fix that. Otherwise, changes are quite well implemented. Good job!

Tharsiga-21:
Thank you.I will check

Tharsiga-21:
@mohdmaazgani I only added the "Contact Support" link and reused the same styling (text-primary hover:underline) that is already used by the existing "Open a GitHub Issue" link. I wasn't able to identify a dark-mode-specific issue caused by this change. Could you please share a screenshot or provide more details about what isn't working in dark mode so I can investigate further?

mohdmaazgani:
@Tharsiga-21 There you go...

https://github.com/user-attachments/assets/38eee4b2-3800-4031-90f5-daa3661125e5

It's working fine with light mode but not in dark mode.
As you can see, when on clicking on the 'Contact Us' section, it should open in the mode (Light or dark) that was initially used in the current website by the user.

Tharsiga-21:
@mohdmaazgani  
Fixed the dark mode issue by adding a useEffect in Contact.tsx that reads the saved theme from localStorage on mount and applies it to the document. The /contact page will now respect the user's current theme when navigated to from the Privacy Policy.

mohdmaazgani:
@Tharsiga-21 Have you checked it thoroughly? It is the same as it was earlier.

Tharsiga-21:
@mohdmaazgani Yeah I checked It works on the netlify deploy for me

Tharsiga-21:
https://github.com/user-attachments/assets/0891fca1-56fe-4cf5-968d-b6a45b728058

mohdmaazgani:
https://github.com/user-attachments/assets/687b3cc4-30bd-48a4-8a7a-a1397f5c3878

@Tharsiga-21 **see this**

mohdmaazgani:
@Tharsiga-21 Any updates or clarification?

mohdmaazgani:
@Tharsiga-21 have you tested this PR on your PC ?, because in my system it is same as it was (theme)

mohdmaazgani:
@Tharsiga-21 Thank you for the contribution.

While reviewing and testing this PR, I noticed a potential issue related to theme persistence. The newly added `Contact Support` link uses:

```tsx
<a href="/contact">
```

When navigating through this link, the application appears to reset to the default light theme, even if the user is currently using dark mode.

Could you please investigate the following points:

1. Check whether using `<a href="/contact">` is causing a full page reload and resetting the application state.
2. If the project uses React Router, consider using:

   ```tsx
   <Link to="/contact">
   ```

   instead, as it preserves client-side state during navigation.
3. Verify that the Contact page correctly inherits the current theme and does not force a light theme through hardcoded styles or theme settings.
4. Ensure that theme preferences are persisted across navigation and page refreshes.
5. Test the following scenarios:

   * Light Mode → Contact Page
   * Dark Mode → Contact Page
   * Refresh on Contact Page while in Dark Mode
   * Production/Live Deployment behavior

Please implement whichever approach best resolves the issue and attach before/after screenshots or a short screen recording demonstrating that the theme remains consistent across navigation.

Thank you for taking a look at this.

Tharsiga-21:
@mohdmaazgani  Thank you for the detailed review.

## Changes Implemented

**1. Replaced `<a href="/contact">` with `<Link to="/contact">`**
The anchor tag was causing a full page reload on navigation, resetting the theme state. Using React Router's `<Link>` keeps navigation client-side and preserves the theme.

**2. Fixed `localStorage` key in `index.html`**
The existing inline script was reading `localStorage.getItem('theme')` but the app stores the theme under `'symptom-scribe-theme'`. This mismatch has been corrected and the script now also handles the `system` theme case.

## Test Results

| Scenario | Result |
|---|---|
| Light Mode → Contact Page (via Link) | ✅ Theme preserved |
| Dark Mode → Contact Page (via Link) | ✅ Theme preserved |
| Refresh on Contact Page in Dark Mode | ⚠️ Appears to work  |
| Refresh on Contact Page in Light Mode | ❌ Theme doesn't persist |

## Additional Finding

The refresh behavior is caused by a **pre-existing bug** — the theme toggle is not saving to `localStorage` at all (returns `null` even after toggling). Because of this, the inline script in `index.html` always falls back to `window.matchMedia('prefers-color-scheme: dark')` — the OS preference.

This means dark mode refresh only *appears* to work because the test device OS is set to dark. It is not actually persisting the user's in-app preference. If a user's OS is dark but they switched the app to light, refresh would still show dark. Both cases are broken — dark just gets lucky due to the OS setting matching.

This appears to be a pre-existing issue on the main branch. Would it be okay if I open a separate PR to fix the theme toggle persistence properly?

Let me know if you need anything else. Thanks!

mohdmaazgani:
Thank you for the thorough investigation and testing.

I appreciate the detailed breakdown of the changes and the test results. The fixes for the navigation issue and the localStorage key mismatch look reasonable, and your explanation regarding the refresh behavior makes sense.

Since the theme persistence issue appears to be a pre-existing bug and is outside the scope of this PR, I am okay with keeping the current changes focused on the Contact Support navigation fix.

Could you please open a separate issue (or PR) specifically for the theme persistence problem and link it here for tracking purposes? That way we can address it independently without mixing concerns.

Thank you for identifying the root cause and documenting the behavior clearly.


> @mohdmaazgani Thank you for the detailed review.
> 
> ## Changes Implemented
> **1. Replaced `<a href="/contact">` with `<Link to="/contact">`** The anchor tag was causing a full page reload on navigation, resetting the theme state. Using React Router's `<Link>` keeps navigation client-side and preserves the theme.
> 
> **2. Fixed `localStorage` key in `index.html`** The existing inline script was reading `localStorage.getItem('theme')` but the app stores the theme under `'symptom-scribe-theme'`. This mismatch has been corrected and the script now also handles the `system` theme case.
> 
> ## Test Results
> Scenario	Result
> Light Mode → Contact Page (via Link)	✅ Theme preserved
> Dark Mode → Contact Page (via Link)	✅ Theme preserved
> Refresh on Contact Page in Dark Mode	⚠️ Appears to work
> Refresh on Contact Page in Light Mode	❌ Theme doesn't persist
> ## Additional Finding
> The refresh behavior is caused by a **pre-existing bug** — the theme toggle is not saving to `localStorage` at all (returns `null` even after toggling). Because of this, the inline script in `index.html` always falls back to `window.matchMedia('prefers-color-scheme: dark')` — the OS preference.
> 
> This means dark mode refresh only _appears_ to work because the test device OS is set to dark. It is not actually persisting the user's in-app preference. If a user's OS is dark but they switched the app to light, refresh would still show dark. Both cases are broken — dark just gets lucky due to the OS setting matching.
> 
> This appears to be a pre-existing issue on the main branch. Would it be okay if I open a separate PR to fix the theme toggle persistence properly?
> 
> Let me know if you need anything else. Thanks!

mohdmaazgani:
@Tharsiga-21 Thank you for the investigation. Since this appears to be a broader theme persistence issue, please create a new issue addressing it and link it here for tracking.

The issue should cover:

* Theme preference persistence across refreshes
* Synchronization between `next-themes` and the custom theme toggler
* Correct behavior of Light, Dark, and System themes
* Verification across navigation and page refreshes

Thank you!

Tharsiga-21:
@mohdmaazgani 
Thanks for the feedback. I've created a separate issue for the theme persistence problem: #396

PR #430, opened 2026-06-19
[Fix]: Implement per-user PBKDF2 salt for key derivation

### LINKED ISSUE #423: [BUG] : PBKDF2 key derivation uses hardcoded global salts
## 🐛 Bug Description
In `src/lib/encryption.ts`, both `deriveKeyFromToken()` and `deriveSearchKeyFromToken()` use hardcoded string salts:
```ts
const salt = encoder.encode("symptom-scribe-offline-salt");
// and
const salt = encoder.encode("symptom-scribe-search-salt");
```
PBKDF2 salts must be unique per credential to prevent precomputation attacks. With a global static salt, an attacker who extracts one user's IndexedDB data can precompute a rainbow table against the known salt and attack all users' encrypted offline records simultaneously.

## 🔄 Steps to Reproduce
1. Log in as any user.
2. Submit symptoms — the record is encrypted with a key derived from `token + "symptom-scribe-offline-salt"`.
3. Extract the IndexedDB entry from DevTools → Application → IndexedDB → SymptomScribeOfflineDB.
4. **Expected Outcome:** The salt used for key derivation is unique to this user, making precomputation infeasible.
5. **Actual Outcome:** The salt is identical for every user in every installation, so a single precomputed table covers all users.

## ✅ Expected Behavior
A random 16-byte salt should be generated once per user on first login and stored in `localStorage` keyed by user ID (e.g. `symptom_scribe_pbkdf2_salt_{userId}`). Subsequent key derivations for that user reuse the stored salt. On logout, the salt entry should be cleared alongside the token.

## 📌 System Information
- **Operating System:** Any
- **Browser/Device:** Any (Chromium recommended for SubtleCrypto + IndexedDB inspection)

## 💡 Additional Context (if any)
The fallback to the old hardcoded salt (when `userId` is not available) is kept to avoid breaking unauthenticated derivation paths. The fix is backward-compatible — existing encrypted records will need one re-encryption cycle on first login after the change, which the existing `onTokenRefresh` hook in `offline-db.ts` already handles. Also export `clearUserSalt(userId)` and call it on logout cleanup.

GirlScript Summer of Code 2026

ishitaajain22-tech:

Added `getUserSalt(userId)` which generates a 16-byte random salt once per user and persists it to localStorage under `symptom_scribe_pbkdf2_salt_{userId`}. Also exports` clearUserSalt(userId)` for logout cleanup. Both` deriveKeyFromToken` and `deriveSearchKeyFromToken` now accept an optional userId parameter and use the per-user salt when provided, falling back to the old domain strings only for unauthenticated paths. `handleSessionChange` extracts `session.user?.id `and passes it to both derive calls.

---

### **1. Type of Change**
- [ ] **Bug Fix** (non-breaking change which fixes an issue)
- [ X] **Refactoring / Performance** (non-breaking code improvements)

---

### **2. Related Issue**
- Fixes #423 

---

- [X ] **Local Build**: The application builds successfully locally (`npm run build`).
- [X ] **Lint/Format Checks**: Local checks passed (`npm run lint` / `npm run format:check`).

---

- [ X] My code follows the code style and formatting guidelines of this project.
- [ X] I have performed a self-review of my own code.
- [ X] I have commented my code, particularly in hard-to-understand areas.
- [ X] My changes generate no new warnings or console errors.
- [ X] There is no commented-out code or debug logs left in the codebase.

### DIFF
--- a/src/lib/encryption.ts
+++ b/src/lib/encryption.ts
@@ -68,8 +68,34 @@ function hexToUint8Array(hex: string): Uint8Array {
   return new Uint8Array(pairs.map((h) => parseInt(h, 16)));
 }
 
+// ─── Per-user PBKDF2 Salt ────────────────────────────────────────────────────
+// A random 16-byte salt is generated once per user and stored in localStorage
+// keyed by user ID. This prevents cross-user precomputation attacks that would
+// be possible with a hardcoded global salt.
+
+const SALT_KEY_PREFIX = "symptom_scribe_pbkdf2_salt_";
+
+function getUserSalt(userId: string): Uint8Array {
+  const storageKey = SALT_KEY_PREFIX + userId;
+  const stored = localStorage.getItem(storageKey);
+  if (stored) {
+    const pairs = stored.match(/[\da-f]{2}/gi) || [];
+    return new Uint8Array(pairs.map((h) => parseInt(h, 16)));
+  }
+  const newSalt = crypto.getRandomValues(new Uint8Array(16));
+  const hex = Array.from(newSalt)
+    .map((b) => b.toString(16).padStart(2, "0"))
+    .join("");
+  localStorage.setItem(storageKey, hex);
+  return newSalt;
+}
+
+export function clearUserSalt(userId: string): void {
+  localStorage.removeItem(SALT_KEY_PREFIX + userId);
+}
+
 // Key Derivation
-export async function deriveKeyFromToken(token: string): Promise<CryptoKey> {
+export async function deriveKeyFromToken(token: string, userId?: string): Promise<CryptoKey> {
   const encoder = new TextEncoder();
   const tokenBytes = encoder.encode(token);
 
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
Hi @ishitaajain22-tech,

Thank you for the security improvement. I reviewed the PR and it looks like it addresses the reported issue by replacing the hardcoded PBKDF2 salts with per-user random salts and wiring the user ID into the key derivation flow.

Before I merge, I would like to clarify two points:

1. I can see that `clearUserSalt(userId)` was added, but I don't see where it is actually invoked during logout. Could you please confirm whether the salt is being removed when a user logs out and point me to the relevant code path?

2. Have you tested this change with users who already have encrypted offline data stored in IndexedDB? Since the key derivation salt changes from the previous static salt to a per-user salt, I want to confirm whether existing offline records remain accessible or if there is a migration/re-encryption strategy in place.

Once I have clarification on these points, I'll be able to make a final decision on the PR.

Thanks for the work!

ishitaajain22-tech:
@mohdmaazgani Hello,
Thank you for the review!
1. Apologies for that, I had missed it. I've pushed the fix to the same PR. handleSessionClear() now calls supabase.auth.getUser() at the top to capture the userId before nulling keys, then calls clearUserSalt(user.id). This covers all logout paths including session expiry since both route through handleSessionClear via onAuthStateChange.
2. This is a known tradeoff of moving away from the static salt, not a regression. Existing offline records won't decrypt with the new per-user derived key — but since offline data is just a local cache of what's already in Supabase, the existing fallback in offline-db.ts handles it cleanly: both tables get cleared and re-sync from the server on next load. Would you prefer I add an explicit one-time migration flag to make that wipe proactive and silent rather than error-triggered? 
Whatever you say

PR #550, opened 2026-07-03
Enhance check-in logic with daily check validation

### LINKED ISSUE #537: [BUG]: Gamification streak check-in can be farmed multiple times per day
## 🐛 Bug Description
`useCheckInChallenge` only guards against duplicate check-ins using a 48-hour window (`now.getTime() - lastCheckIn.getTime() < 48 * 60 * 60 * 1000`) instead of checking whether the user already checked in **today** (calendar-day comparison). Combined with the fact the read (`select`) and write (`update`) are two separate, non-atomic requests, a user can call the check-in mutation repeatedly (e.g. every few minutes) and the streak counter (`current_streak`) will increment on every call as long as each call happens within 48 hours of the previous one.
 
## 🔄 Steps to Reproduce
1. **Go to:** the Challenges tab and join any challenge.
2. **Click on:** "Check in" twice (or more) within the same day, or via repeated API calls.
3. **Expected Outcome:** Only one check-in should be accepted per calendar day; the streak should increase by at most 1 per day.

## 📌 System Information
- **Operating System:** Any
- **Browser/Device:** Any

 
## 🏆 Are you contributing under an open-source program? ggsoc

ishitaajain22-tech:
**File:** `src/hooks/useGamification.ts`
 
`useCheckInChallenge` only rejected a check-in if the previous one was **less than 48 hours** ago, not "already checked in today." It also did a client-side fetch-then-update across two separate requests, so it wasn't atomic. A user could call the mutation repeatedly and inflate `current_streak` multiple times in a single day.
 
---

### **1. Type of Change**

---

### **2. Related Issue**
- Fixes #537 

---

- [ X] **Local Build**: The application builds successfully locally (`npm run build`).
- [ X] **Lint/Format Checks**: Local checks passed (`npm run lint` / `npm run format:check`).
- [ X] **Manual Verification**: Briefly list the steps/features verified manually.

---

- [ X] I have performed a self-review of my own code.
- [ X] I have commented my code, particularly in hard-to-understand areas.
- [ X] My changes generate no new warnings or console errors.
- [ X] There is no commented-out code or debug logs left in the codebase.

### DIFF
--- a/src/hooks/useGamification.ts
+++ b/src/hooks/useGamification.ts
@@ -139,7 +139,7 @@ export function useCheckInChallenge() {
       if (fetchErr) throw fetchErr;
 
       const now = new Date();
-      const todayDateStr = now.toISOString().split("T")[0]; // e.g. "2026-07-03"
+      const todayDateStr = now.toISOString().split("T")[0]; 
       const todayStartISO = `${todayDateStr}T00:00:00.000Z`;
 
       const lastCheckIn = current.last_check_in ? new Date(current.last_check_in) : null;
@@ -150,8 +150,16 @@ export function useCheckInChallenge() {
         return { streak: current.current_streak, alreadyCheckedInToday: true };
       }
 
-      const isConsecutive =
-        lastCheckIn !== null && now.getTime() - lastCheckIn.getTime() < 48 * 60 * 60 * 1000;
+      const lastCheckInDateStr = lastCheckIn ? lastCheckIn.toISOString().split("T")[0] : null;
+      const daysSinceLastCheckIn = lastCheckInDateStr
+        ? Math.round(
+            (Date.parse(`${todayDateStr}T00:00:00.000Z`) -
+              Date.parse(`${lastCheckInDateStr}T00:00:00.000Z`)) /
+              (24 * 60 * 60 * 1000)
+          )
+        : null;
+
+      const isConsecutive = daysSinceLastCheckIn === 1;
       const newStreak = isConsecutive ? current.current_streak + 1 : 1;
 
       const { data: updated, error } = await supabase
@@ -169,8 +177,6 @@ export function useCheckInChallenge() {
       if (error) throw error;
 
       if (!updated) {
-        // Someone (or another tab/request) already checked in today between
-        // our read and this write — don't double-count the streak.
         return { streak: current.current_streak, alreadyCheckedInToday: true };
       }
 
--- a/src/hooks/useGamification.ts
+++ b/src/hooks/useGamification.ts
@@ -131,7 +131,6 @@ export function useCheckInChallenge() {
   const qc = useQueryClient();
   return useMutation({
     mutationFn: async (userChallengeId: string) => {
-      // Get current record
       const { data: current, error: fetchErr } = await supabase
         .from("user_challenges")
         .select("*")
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
Thanks for working on this fix. I reviewed the implementation and have a few questions before I can approve the PR:



1. **Timezone Handling**
   - The daily validation is based on:
     ```ts
     now.toISOString().split("T")[0]
     ```
     which uses UTC dates. Is the intended behavior to enforce check-ins based on UTC days or the user's local calendar day?
   - Have you tested the flow around midnight for users in non-UTC timezones?

2. **Streak Calculation Logic**
   - The PR introduces daily check-in validation, but the streak continuation logic still relies on:
     ```ts
     now.getTime() - lastCheckIn.getTime() < 48 * 60 * 60 * 1000
     ```
   - Should streaks be determined by consecutive calendar days instead? For example, if a user misses an entire day but remains within a 48-hour window, should the streak continue or reset?

3. **Concurrency / Atomicity**
   - The PR description mentions that the previous implementation was not atomic. While the conditional update significantly reduces the race window, the flow still performs a read followed by an update.
   - Can you explain how this approach guarantees correctness under concurrent requests, and whether you considered moving the check-in logic to a database function/RPC for true atomic execution?

4. **Validation Coverage**
   - Have you tested the following scenarios?
     - Multiple check-ins from the same browser tab
     - Multiple tabs opened simultaneously
     - Rapid repeated clicks on the check-in action
     - Midnight boundary transitions

5. **Security / Data Ownership**
   - Is access to `user_challenges` protected through RLS policies such that a user cannot update another user's challenge record by supplying a different `userChallengeId`?

Once these points are clarified, I'll be able to complete the review.

ishitaajain22-tech:
@mohdmaazgani hello

Thanks for the thorough review , a couple of these pointed at a real bug, so I've pushed a fix rather than just answering around it. Going through each:

**1\. Timezone handling**Yes, the daily boundary is UTC, not the user's local calendar day. I haven't tested this against a non-UTC client around midnight — the logic itself is deterministic and correct _for UTC_, but a user in, say, IST (UTC+5:30) will see their "day" roll over at 5:30 AM local time, not midnight local. I don't think this breaks functionality (the streak logic is still internally consistent), but it is a real UX rough edge. I'd suggest treating this as a known, accepted limitation for this PR and filing a follow-up if we want true local-day boundaries — that needs the client's timezone (or offset) sent to the server, which is a bigger change than this fix.

**2\. Streak calculation logic** This was a genuine inconsistency, not just an edge case. I added the same-day guard, but left the "is this a continuing streak" check as the old 48-hour rolling window. Your example is exactly right: a user checking in at 23:59 on day 1 and 00:01 on day 3 is within 48 hours but skipped an entire calendar day. I've fixed it to compare UTC date strings directly — the streak only continues if daysSinceLastCheckIn === 1; any gap of 2+ calendar days resets it to 1. This is now consistent with the same-day guard rather than using a different time window for each check.

**3\. Concurrency / atomicity**The conditional update does give correctness here, not just a "reduced window" — here's why: the UPDATE ... WHERE ... last\_check\_in IS NULL OR last\_check\_in < today runs as a single statement. Postgres takes a row lock on UPDATE; if two requests race, the second one blocks until the first commits, then **re-evaluates its own WHERE clause against the now-committed row** — at that point last\_check\_in is today's date, so the second update's WHERE no longer matches and it affects 0 rows. That's what .maybeSingle() returning null detects. The one part that isn't fully atomic is the initial SELECT used to compute newStreak — but since both concurrent requests would read the same pre-write state and compute the same newStreak, and only one of them can ever actually commit a write, there's no scenario where the streak gets double-incremented. That said, I want to be upfront: this reasoning is correct under Postgres's default READ COMMITTED isolation, but it's less generalizable than an RPC would be if this logic grows more complex later (e.g. if we ever need to touch multiple rows atomically as part of check-in). I'd treat "move to an RPC" as a reasonable follow-up if we want stronger guarantees going forward, not something blocking this PR.

**4\. Validation coverage**Same-tab double-click and rapid repeated clicks: covered — the second call's conditional UPDATE matches 0 rows and returns alreadyCheckedInToday: true. Multiple tabs / concurrent requests: covered by the same mechanism (see 3). Midnight boundary: **not yet manually tested** with a real clock crossing UTC midnight — I've reasoned through it via the date-string comparison, but I'll add an explicit test (mocking Date around a UTC day boundary) before merge rather than just asserting it works.

**5\. Security / data ownership**I want to flag this honestly rather than guess: the update query filters only on .eq("id", userChallengeId) — there's no .eq("user\_id", ...) in the query itself, so ownership enforcement is entirely dependent on RLS at the database level. I checked every migration in this repo and couldn't find a CREATE TABLE or CREATE POLICY for user\_challenges anywhere — it isn't in 20260609144200\_enforce\_rls\_policies.sql alongside profiles/symptom\_history/health\_metrics/chat\_sessions, and I don't see it created anywhere else either. So I can't confirm from the codebase whether it's RLS-protected. Could you point me to where that table/policy is defined, or should this PR add an explicit RLS policy (and a defense-in-depth .eq("user\_id", user.id) in the query) as part of this fix? I'd rather not merge this assuming RLS exists if nobody can point to where it's enforced.

PR #552, opened 2026-07-03
Refactor Gemini API integration for streaming response

### LINKED ISSUE #540: [BUG] : `symptom-analyzer` fakes SSE streaming (fetches full response, then sends one chunk)
## 🐛 Bug Description
The function sets `Content-Type: text/event-stream`, `Cache-Control: no-cache`, `Connection: keep-alive`, and formats the response as an SSE stream — implying token-by-token streaming. In reality it calls Gemini's non-streaming `generateContent` endpoint, `await`s the **entire** response, and only then wraps the whole finished text into a single `data: {...}` SSE event followed by `data: [DONE]`. There is no real streaming; the client waits for the full generation time with zero intermediate output, while still paying the overhead/complexity of an SSE parser on the frontend.
 
## 🔄 Steps to Reproduce
1. **Go to:** the AI Symptom Checker chat.
2. **Click on:** Send a symptom description and observe the response area.
3. **Expected Outcome:** Text should appear incrementally as the model generates it (as the `text/event-stream` content type and streaming UI imply).
4. **Actual Outcome:** Nothing appears until the entire Gemini response is ready, then the full answer appears all at once — functionally a blocking request dressed up as a stream.
 
## 📌 System Information
- **Operating System:** N/A (server-side)
- **Browser/Device:** Any

ishitaajain22-tech:
**File:** `supabase/functions/symptom-analyzer/index.ts`
 
The function set `Content-Type: text/event-stream` and framed the response as SSE, but called Gemini's non-streaming `generateContent` endpoint, awaited the **full** response, then wrapped the entire finished answer into a single `data:` event. Users saw nothing until the whole generation finished — the opposite of what an SSE endpoint implies.

---

### **1. Type of Change**

---

### **2. Related Issue**
- Fixes #540 

---

- [X] **Manual Verification**: Briefly list the steps/features verified manually.

---

### DIFF
--- a/supabase/functions/symptom-analyzer/index.ts
+++ b/supabase/functions/symptom-analyzer/index.ts
@@ -10,10 +10,17 @@ const ALLOWED_ORIGINS = [
   "http://localhost:3000",
   "http://localhost:8080",
   "https://symptom-scribe.vercel.app",
+  "https://symptom-scribe-clean.netlify.app",
 ];
 
+const NETLIFY_PREVIEW_ORIGIN = /^https:\/\/deploy-preview-\d+--symptom-scribe-clean\.netlify\.app$/;
+
+function isAllowedOrigin(origin: string): boolean {
+  return ALLOWED_ORIGINS.includes(origin) || NETLIFY_PREVIEW_ORIGIN.test(origin);
+}
+
 const getCorsHeaders = (origin: string | null) => ({
-  "Access-Control-Allow-Origin": origin && ALLOWED_ORIGINS.includes(origin) ? origin : "null",
+  "Access-Control-Allow-Origin": origin && isAllowedOrigin(origin) ? origin : "null",
   "Access-Control-Allow-Headers": "authorization, x-client-info, apikey, content-type",
 });
 
@@ -21,7 +28,7 @@ serve(async (req: Request): Promise<Response> => {
   const origin = req.headers.get("origin");
 
   // 1. Origin allowlist check
-  if (origin && !ALLOWED_ORIGINS.includes(origin)) {
+  if (origin && !isAllowedOrigin(origin)) {
     return jsonResponse({ error: "Origin not allowed" }, 403, getCorsHeaders(origin));
   }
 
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
image

mohdmaazgani:
1. Why was `https://symptom-scribe-clean.netlify.app` removed from `ALLOWED_ORIGINS`?
   - This change appears unrelated to Issue #540.
   - Could this break Netlify deployments or previews?

2. Have you verified the exact response format returned by Gemini's
   `streamGenerateContent?alt=sse` endpoint?
   - Can you provide a sample streamed event payload?

3. What happens if the client disconnects while Gemini is still streaming?
   - Should the upstream reader be cancelled?

4. At stream completion, could there still be an unprocessed JSON fragment
   remaining in `buffer`?
   - Have you tested the final-buffer edge case?

5. If a streaming error occurs midway through generation,
   should the client receive an explicit error event instead of only `[DONE]`?

6. Have you manually verified that the frontend now renders content incrementally
   rather than waiting for the full response?
   - A short screen recording or test evidence would be helpful.

ishitaajain22-tech:
Thanks for the thorough review. Sorry for the points I missed

**1\. Why was symptom-scribe-clean.netlify.app involved, and could it break deploys?**

Removed my mistake. Fixed by adding the production Netlify domain to the allowlist and matching deploy previews with a pattern instead of requiring an exact string match:

ts
`   const NETLIFY_PREVIEW_ORIGIN = /^https:\/\/deploy-preview-\d+--symptom-scribe-clean\.netlify\.app$/;  function isAllowedOrigin(origin: string): boolean {    return ALLOWED_ORIGINS.includes(origin) || NETLIFY_PREVIEW_ORIGIN.test(origin);  }   `

This should be applied to the other edge functions too for consistency — I only touched symptom-analyzer since that's what's in scope for #540. Happy to do the rest as a follow-up commit if you'd like it bundled here instead.

**2\. Have you verified Gemini's actual streamGenerateContent?alt=sse response format?**

Yes — verified against Google's own streaming docs and quickstart examples rather than assuming. Each SSE event is data: {...}\\n\\n where the JSON is a full GenerateContentResponse object, e.g.:

`   data: {"candidates":[{"content":{"parts":[{"text":"partial "}],"role":"model"},"index":0}],"usageMetadata":{...}}   `

The text field is the incremental delta, not the accumulated text. One thing my first pass missed: parts can contain **more than one** text entry in a single event, and I was only reading parts\[0\]. Fixed to iterate all parts and emit each one.

**3\. What happens if the client disconnects mid-stream — is the upstream reader cancelled?**

It wasn't before — good catch. Added a cancel() handler on the ReadableStream's underlying source, so when the client disconnects (tab closed, navigated away, fetch aborted), we cancel the upstream Gemini reader too instead of continuing to read/pay for a full generation nobody is listening to anymore.

**4\. Could an unprocessed JSON fragment remain in buffer at stream completion — was this tested?**

Yes, this was a real gap. If the upstream body ends without a trailing newline after the final event, that last line was sitting in buffer and was never processed — silently dropping the tail of the answer. Fixed by flushing and processing whatever remains in buffer after the read loop exits, before sending \[DONE\].

**5\. Should a mid-stream error produce an explicit error event instead of only \[DONE\]?**

Agreed, and fixed — previously a mid-stream failure just fell through to \[DONE\] with no way for the client to distinguish it from a clean finish. Now an explicit event is sent first:

`   data: {"error": "Stream interrupted while generating the response."}   `

followed by \[DONE\]. 

**6\. Have you manually verified incremental rendering — can you provide a recording?**

I've fixed the server-side logic so it forwards real deltas as they arrive instead of buffering the full response. I'll test it in-browser and get you a short recording before asking for approval — will follow up with that separately.

mohdmaazgani:
The streaming implementation looks good and addresses the root cause of Issue #540.

However, I noticed that `reader` is declared inside `start()` but referenced inside `cancel()`:

```ts
async start(controller) {
  const reader = geminiResponse.body!.getReader();
}

cancel(reason) {
  reader.cancel(reason);
}
```
Please check and have a look at it.

ishitaajain22-tech:
@mohdmaazgani done as per your request.

PR #553, opened 2026-07-03
Refactor health metrics sync logic

### LINKED ISSUE #541: [BUG] : `get-cached-data` returns stale data that silently overwrites
## 🐛 Bug Description
`fetchHistory` fetches from `getCachedData("health_metrics")` (which may return a Redis cache entry up to 300s / 5 minutes old — see `get-cached-data/index.ts` TTL), then unconditionally **deletes all locally-synced records** (`pending_sync === 0`) for the user and replaces them with whatever the cache returned. If a record was added from another device/tab within the last 5 minutes, the stale cache response will not contain it, and this function will delete the local (correct, freshly-synced) copy and replace it with the stale set — causing a real, recently-saved health metric to disappear from the UI until the cache naturally expires.
 
## 🔄 Steps to Reproduce
1. **Go to:** Health Metrics page on Device/Tab A. Add a new metric (it syncs to Supabase, invalidating nothing on Device B's cache).
2. **Click on:** Refresh Health Metrics page on Device/Tab B within 5 minutes of the addition (or same tab, right after another mutation path that doesn't call `invalidateCache`).
3. **Expected Outcome:** The newly added metric shows up, or at minimum previously-synced local data is preserved.
4. **Actual Outcome:** The stale cached list (missing the new metric) overwrites the local IndexedDB copy of synced records, so the metric appears to vanish.

## 📌 System Information
- **Operating System:** Windows/macOS/Linux (any, since this is an offline-first IndexedDB flow)
- **Browser/Device:** Any
## 💡 Additional Context (if any)
This is a health-tracking app, so silently dropping metric entries from the visible history (even temporarily) is a meaningful data-trust bug.

ishitaajain22-tech:
 
**File:** `src/hooks/useMetricsHistory.ts`
 
On every online fetch, the hook deleted **all** locally-synced `health_metrics` rows for the user and replaced them with whatever `get-cached-data` returned — a Redis cache entry that can be up to 5 minutes stale (see the function's `ex: 300` TTL). If a metric was added on another device/tab within that window, the stale response wouldn't contain it, and this delete-then-replace wiped the locally-synced, correct copy.

---

### **1. Type of Change**

---

### **2. Related Issue**
- Fixes #541 

---

- [X] **Manual Verification**: Briefly list the steps/features verified manually.

---

### DIFF
--- a/src/hooks/useMetricsHistory.ts
+++ b/src/hooks/useMetricsHistory.ts
@@ -19,7 +19,7 @@ export function useMetricsHistory(userId: string | null) {
 
       if (navigator.onLine) {
         try {
-          const { data, error } = await getCachedData<OfflineMetric[]>("health_metrics");
+          const { data, cachedAt, error } = await getCachedData<OfflineMetric[]>("health_metrics");
 
           if (!error && data) {
             const localEntries = data.map((record) => ({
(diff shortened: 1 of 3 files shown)


mohdmaazgani:
Thanks for the fix and the detailed explanation. Before I can approve this PR, I have one concern regarding the cleanup logic:

```ts
.filter((record) => new Date(record.recorded_at).getTime() < staleCutoff)
```

From my understanding, `recorded_at` represents when the health metric was recorded, not when the cache was generated or last refreshed.

Could you clarify why `recorded_at` is being used as a proxy for cache staleness?

For example, a metric recorded several days ago can still be valid data. If that metric is temporarily absent from the cached response, would this logic make it eligible for deletion simply because its `recorded_at` timestamp is older than the 5-minute cache window?

Also, is there a reason the cache TTL is hardcoded here:

```ts
const CACHE_STALENESS_MS = 5 * 60 * 1000;
```

Would it be safer to derive this from a shared configuration so frontend and backend cache behavior cannot drift over time?

ishitaajain22-tech:
@mohdmaazgani Hi
Good catch, thanks for pushing on this. You're right on both counts, and I've pushed a fix for the real issue rather than just the symptom.
On recorded_at as a proxy: it doesn't tell us anything about cache freshness — it's just when the metric happened. The comparison I actually needed wasn't "is this record older than some TTL guess," it was "did this record exist before the specific cache snapshot I'm looking at was generated." Those aren't the same thing, and your example (an old, valid record temporarily missing from one snapshot for an unrelated reason) exposes exactly where they diverge.
Fixed it by having `get-cached-data` return the snapshot's own generation time (cachedAt) alongside the data, instead of the client inferring anything. Now a local record only gets removed if recorded_at` < cachedAt` — i.e. it definitely existed before this snapshot was taken and is still missing from it, which is the only case where "missing from cache" reliably means "deleted server-side." A record newer than the snapshot is never touched, regardless of its absolute age.
On the hardcoded TTL: agreed, and this removes the need for it entirely — the client no longer has to know or guess the Redis TTL at all, so there's nothing left that can drift between the two sides. Updated `getCachedData() `in cached-queries.ts to unwrap` { data, cachedAt }`; existing call sites` (Dashboard, History) `are unaffected since they only destructure `{ data, error }`.
I have pushed the new changes

PR #559, opened 2026-07-04
feat(dashboard): add personalized welcome message

### LINKED ISSUE #546: [Feature] Add personalized welcome message on Health Dashboard
## ✨ Feature Overview
Display a personalized welcome message above the Health Dashboard whenever a user logs in.
Example:
- Welcome, John!
This will make the dashboard feel more personalized and improve the overall user experience.

## 🚀 Why is this Feature Needed?
Currently, users land directly on the dashboard after authentication. Adding a personalized greeting creates a warmer first impression and helps users feel more connected to the application.

## 💡 Possible Implementation (Optional)

- Fetch the authenticated user's name from the current session.
- Display a greeting above the "Health Dashboard" heading.
- Show the greeting only after a successful login.
- Ensure proper responsiveness and compatibility with both light and dark themes.

priya434960:
- Fixes #546

Screenshot 2026-07-04 145608

### DIFF
--- a/src/pages/Dashboard/Dashboard.test.tsx
+++ b/src/pages/Dashboard/Dashboard.test.tsx
@@ -100,6 +100,17 @@ function mockCachedSymptoms(data: unknown[] | null, error: unknown = null) {
 function mockAuthUser(user: typeof mockUser | null = mockUser) {
   (supabase.auth.getUser as Mock).mockResolvedValue({ data: { user } });
 }
+(supabase.from as Mock).mockReturnValue({
+    select: () => ({
+      eq: () => ({
+        maybeSingle: () =>
+          Promise.resolve({
+            data: { full_name: "User" },
+          }),
+      }),
+    }),
+  });
+
 
 // ---------------------------------------------------------------------------
 // Sample fixture data
--- a/src/pages/Dashboard/index.tsx
+++ b/src/pages/Dashboard/index.tsx
@@ -151,6 +151,7 @@ const Dashboard = () => {
   });
   const [recentHistory, setRecentHistory] = useState<SymptomHistoryRecord[]>([]);
   const [loading, setLoading] = useState(true);
+  const [userName, setUserName] = useState("User");
 
   useEffect(() => {
     fetchDashboardData();
@@ -161,6 +162,15 @@ const Dashboard = () => {
       const {
         data: { user },
       } = await supabase.auth.getUser();
+      const { data: profile } = await supabase
+        .from("profiles")
+        .select("full_name")
+        .eq("user_id", user.id)
+        .maybeSingle();
+
+      if (profile?.full_name) {
+        setUserName(profile.full_name);
+      }
       if (!user) {
         setLoading(false);
         return;
@@ -260,6 +270,22 @@ const Dashboard = () => {
 
   return (
     <div className="space-y-6">
+      <Card className="border bg-gradient-to-r from-blue-50 to-white dark:from-slate-900 dark:to-slate-800">
+        <CardContent className="flex items-center justify-between py-6">
+          <div>
+            <h2 className="text-3xl font-bold">
+              Welcome, <span className="text-blue-600">{userName}</span>
+            </h2>
+            <p className="text-muted-foreground mt-1">
+              Advanced analytics & personalized health insights
+            </p>
+          </div>
+
+          <div className="w-14 h-14 rounded-full bg-blue-100 flex items-center justify-center">
+            <Activity className="h-7 w-7 text-blue-600" />
+          </div>
+        </CardContent>
+      </Card>
       <div>
         <h1 className="text-3xl font-bold text-foreground">Health Dashboard</h1>
         <p className="text-muted-foreground">Overview of your health tracking journey</p>
--- a/src/pages/Dashboard/index.tsx
+++ b/src/pages/Dashboard/index.tsx
@@ -9,7 +9,7 @@ import CountUp from "react-countup";
 import CardSkeleton from "@/components/ui/CardSkeleton";
 import { getCachedData } from "@/lib/cached-queries";
 import { decryptSymptom, type OfflineSymptom } from "@/lib/offline-db";
-import { whenEncryptionReady } from "@/lib/encryption";
+import { whenEncryptionReady, decryptProfileField } from "@/lib/encryption";
 
 interface Stats {
   totalSymptoms: number;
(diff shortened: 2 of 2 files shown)


mohdmaazgani:
Thanks for the contribution.

After testing the deploy preview, the welcome banner is displaying an internal identifier instead of a user-friendly name:

Welcome,
f44db5486114ef622c9c97e5:d2710ef...

This suggests `profiles.full_name` may not always contain a valid display name. Please verify the source field and add a proper fallback strategy.

Additionally:
- Move the `if (!user)` check before accessing `user.id`.
- Add handling for missing or invalid profile names.
- Consider reducing duplication between the welcome card and the existing dashboard heading.

At the moment the PR is not ready to merge.

Screenshot 2026-07-04 202055

priya434960:
Hi, @mohdmaazgani 
I'm currently working on the encryption changes, but I've run into an issue. The dashboard is failing with OperationError during decryptProfileField(), and I'm also seeing "Failed to load dashboard data." I've verified that the encryption_salt migration has been applied, but the decryption issue still persists. I suspect there may be a mismatch in how the encryption keys are being derived, but I'm not completely sure. Could you please guide me in the right direction or let me know if I'm missing something? I've been debugging this for quite a while, so I'd really appreciate your help. Thank you!

mohdmaazgani:
Hi @priya434960,

Thanks for looking into this.

The `OperationError` from `decryptProfileField()` usually means the decryption key being generated at runtime does not match the key that was originally used to encrypt the data. Since you've already confirmed the `encryption_salt` migration is applied, I would suggest checking the following:

1. Verify that the same encryption logic is being used for both encryption and decryption (salt, key derivation parameters, encoding, etc.).
2. Check whether the affected records were encrypted before the recent encryption changes. Older records may not be compatible with the current implementation.
3. Inspect the values being passed into `decryptProfileField()` and confirm they are valid encrypted payloads and not plain text or partially migrated data.
4. Compare the key derivation path during encryption vs decryption and ensure they produce identical keys for the same user.
5. Check browser console and Supabase logs for the full stack trace to identify exactly which record is triggering the failure.

For the dashboard issue, the "Failed to load dashboard data" message may just be a downstream effect of the decryption failure. I would first focus on identifying why `decryptProfileField()` is throwing and whether a specific profile record is causing it.

Could you share:
- The full error stack trace.
- The exact record/field being decrypted when it fails.
- The changes you've made related to encryption.

That should help narrow down the root cause much faster.

priya434960:
Hi @mohdmaazgani,
I investigated the issue further and here are my findings:
The encryption_salt migration has been applied successfully.
I verified that the profiles table now contains the encryption_salt column.
whenEncryptionReady() completes successfully and a valid CryptoKey is generated.
The encrypted profile value is fetched correctly from the database.
However, decryptProfileField() still fails with OperationError inside decryptText().
I also deleted my existing profiles row and tested again with a fresh profile to rule out old encrypted data, but the issue still persists. The dashboard still shows "Welcome User", and the profile name is not decrypted.
From the browser console, the relevant error is:
Full name decryption failed: OperationError
at decryptText(...)
at decryptProfileField(...)
There are also CORS errors for the get-cached-data Edge Function, but those appear to be separate from the decryption issue.
Could you please advise if there is any additional encryption setup or key/salt synchronization step that I may be missing, or if this could be an issue with the current encryption flow?
Thanks!

mohdmaazgani:
Hi @priya434960,

Thanks for the detailed investigation. Based on what you've checked so far, it sounds like we've ruled out the migration itself and also ruled out legacy encrypted data.

Since:
- `encryption_salt` exists in the database
- `whenEncryptionReady()` completes successfully
- A valid `CryptoKey` is generated
- The encrypted value is fetched correctly
- The issue reproduces even with a freshly created profile

the problem is likely somewhere in the encryption/decryption flow rather than the migration.

Could you please check the following:

### 1. Verify encrypt → decrypt roundtrip locally
Try encrypting a simple test value and immediately decrypting it using the same code path:

```ts
const encrypted = await encryptProfileField("Test User");
const decrypted = await decryptProfileField(encrypted);

console.log({ encrypted, decrypted });
```

If this fails, then the issue is in the encryption implementation itself rather than the database.

### 2. Check whether the same salt is used for both operations
Please confirm that:
- The salt used during encryption
- The salt used during decryption

are identical for the same profile.

A mismatch here would result in the `OperationError` you're seeing.

### 3. Inspect the encrypted payload structure
Please compare:
- The payload returned by `encryptProfileField()`
- The payload passed into `decryptProfileField()`

and verify that no fields (IV, ciphertext, encoding, etc.) are being modified before decryption.

### 4. Add temporary logging
Could you log:
- Generated salt
- IV
- Encrypted payload length
- Decrypted payload input

around `encryptText()` and `decryptText()`?

That should help identify whether the failure is occurring before the Web Crypto API call or inside the actual AES decryption.



At this stage, I suspect either:
- A key derivation mismatch,
- A payload format issue,
- An IV mismatch,
- Or an algorithm/configuration mismatch between encryption and decryption.

The fact that the issue still occurs with a freshly created profile suggests this is more likely a code-level issue than a data migration issue.

Thanks again for digging into this. Once we can see the encryption helpers or additional logs, we should be able to narrow down the root cause much more quickly.

priya434960:
Hi, @mohdmaazgani 
I have pushed an updated version based on my understanding. I am not completely sure if it is the best approach, so I'd really appreciate it if you could review it whenever you have some time.
And it is working correctly in my local environment now.
Thank you!

mohdmaazgani:
@priya434960 There is a major problem. What I just noticed is that it is not compatible with different themes. Make sure to raise the PR to rectify this error as soon as possible

image
image
image
and many more.....

mohdmaazgani:
Hi @priya434960 , are you working on this ?

> @priya434960 There is a major problem. What I just noticed is that it is not compatible with different themes. Make sure to raise the PR to rectify this error as soon as possible
> 
> image image image and many more.....

priya434960:
Hi @mohdmaazgani 
Yes, I am currently working on resolving the issue. I was unable to be active over the last few days due to some health-related reasons. I will provide an update by tonight, or by tomorrow morning at the latest. Thank you for your understanding and patience.

mohdmaazgani:
@priya434960 No worries! and take care.

> Hi @mohdmaazgani Yes, I am currently working on resolving the issue. I was unable to be active over the last few days due to some health-related reasons. I will provide an update by tonight, or by tomorrow morning at the latest. Thank you for your understanding and patience.

PR #561, opened 2026-07-04
[fix]: Permissive CORS fallback on account deletion

### LINKED ISSUE #555: [BUG] : Duplicate/conflicting account-deletion edge functions
**File:** `supabase/functions/delete-account/index.ts`
 
## 🐛 Bug Description
There are two separate edge functions that both delete the currently authenticated user's account: `delete-account/index.ts` and `delete-user-account/index.ts`. They are near-duplicates but not identical — e.g. `delete-account` uses the service-role client only (`createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY)`) and calls `auth.admin.deleteUser(user.id)` without ever validating that dependent user data (symptom history, chat sessions, health metrics not covered by DB-level cascade) is cleaned up first if cascade constraints are missing/out of date. 
 
## 🔄 Steps to Reproduce
1. **Go to:** Account Settings → "Delete Account" in the frontend (uses one of these two functions — check which one is actually wired up on the client).
2. **Click on:** Delete Account.
3. **Expected Outcome:** A single, well-defined, consistently-secured deletion endpoint is called.
4. **Actual Outcome:** Two functions exist with different CORS fallback behavior and no shared source of truth, making it unclear which is authoritative, increasing the attack surface, and risking one of them drifting out of sync with future data-cleanup requirements (e.g. new tables added later only get cleanup logic added to one of the two).

## 📌 System Information
- **Operating System:** N/A (server-side)
- **Browser/Device:** Any
 
ggsoc

ishitaajain22-tech:

**File:** `supabase/functions/delete-account/index.ts`
 
For disallowed origins, `getCorsHeaders` fell back to `ALLOWED_ORIGINS[0]` (`http://localhost:3000`) instead of denying, unlike `get-cached-data`/`invalidate-cache`/`symptom-analyzer`, which return `"null"`. On an irreversible, destructive endpoint (`auth.admin.deleteUser`), this is the wrong endpoint to be more permissive on, not less.

---

### **1. Type of Change**
- [X] **Bug Fix** (non-breaking change which fixes an issue)
- [X] **New Feature** (non-breaking change which adds functionality)
- [X] **Refactoring / Performance** (non-breaking code improvements)
- [X] **Documentation Update** (changes to README, guides, or comments)
- [X] **Other** (please specify): _________

---

### **2. Related Issue**
- Fixes #555 

---

---

- [X] My code follows the code style and formatting guidelines of this project.

### DIFF
--- a/supabase/functions/delete-account/index.ts
+++ b/supabase/functions/delete-account/index.ts
@@ -9,10 +9,13 @@ const ALLOWED_ORIGINS = [
 ];
 
 const getCorsHeaders = (origin: string | null) => ({
+  // Previously fell back to ALLOWED_ORIGINS[0] ("http://localhost:3000") for
+  // any disallowed origin, which meant a preflight from an untrusted origin
+  // could still receive a permissive Access-Control-Allow-Origin header on
+  // this irreversible, destructive endpoint. Deny by default instead,
+  // consistent with get-cached-data / invalidate-cache / symptom-analyzer.
   "Access-Control-Allow-Origin":
-    origin && ALLOWED_ORIGINS.includes(origin)
-      ? origin
-      : ALLOWED_ORIGINS[0],
+    origin && ALLOWED_ORIGINS.includes(origin) ? origin : "null",
   "Access-Control-Allow-Headers":
     "authorization, x-client-info, apikey, content-type",
 });
--- a/supabase/functions/delete-account/index.ts
+++ b/supabase/functions/delete-account/index.ts
@@ -9,11 +9,6 @@ const ALLOWED_ORIGINS = [
 ];
 
 const getCorsHeaders = (origin: string | null) => ({
-  // Previously fell back to ALLOWED_ORIGINS[0] ("http://localhost:3000") for
-  // any disallowed origin, which meant a preflight from an untrusted origin
-  // could still receive a permissive Access-Control-Allow-Origin header on
-  // this irreversible, destructive endpoint. Deny by default instead,
-  // consistent with get-cached-data / invalidate-cache / symptom-analyzer.
   "Access-Control-Allow-Origin":
     origin && ALLOWED_ORIGINS.includes(origin) ? origin : "null",
   "Access-Control-Allow-Headers":
--- a/supabase/functions/delete-account/index.ts
+++ b/supabase/functions/delete-account/index.ts
@@ -6,11 +6,18 @@ const ALLOWED_ORIGINS = [
   "http://localhost:3000",
   "http://localhost:8080",
   "https://symptom-scribe.vercel.app",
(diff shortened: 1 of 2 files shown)


mohdmaazgani:
Thanks for the contribution.

A few clarifications before approval:

1. Issue #555 primarily reports duplicate/conflicting account deletion functions. This PR updates CORS handling only. Can you explain how this resolves the duplication concern?

2. Which account deletion function is currently used by the frontend:
   - delete-account
   - delete-user-account

3. Have you tested account deletion from:
   - Production Netlify site
   - Netlify Deploy Preview
   - localhost

4. Was Access-Control-Allow-Origin: "null" tested in all deletion flows?

5. Are there any remaining CORS differences between the two deletion functions?

ishitaajain22-tech:
@mohdmaazgani hello

Thanks for the review.

\. Which function does the frontend use?

delete-user-account — confirmed at src/pages/User/Settings.tsx (supabase.functions.invoke("delete-user-account")). delete-account exists in the codebase but isn't called from anywhere in the frontend.

\. Does this PR resolve the duplication concern?

Not on its own, and that's a fair thing to flag — the original scope of this PR was the CORS fallback specifically, which is only half of what #555 is asking for. Since confirming which function is actually live (point 2), it makes sense to widen the fix to the function that matters in production rather than leave the CORS improvement sitting on the unused one. I've done that below.

\. Are there remaining CORS differences between the two functions?

Yes — and reviewing delete-user-account closely surfaced a second, independent issue worth fixing at the same time: it calls client.auth.getUser() with no token argument, while the client is only configured with the header via global.headers. That combination doesn't populate a session for getUser() to read from, so it can fail authentication even for a valid JWT. delete-account already passes the token explicitly and doesn't have this issue. I've aligned delete-user-account to the same explicit-token pattern as part of this update, along with the CORS deny-by-default fix and adding the Netlify origins.

Testing across Production / Preview / localhost, and the "null" fallback in all flows

Tested locally on localhost:8080 and the auth/CORS fix works as expected there.
I don't have access to trigger or view the Netlify deploy preview or production site myself — that needs either a PR opened against the connected branch (which auto-generates the preview) or someone with dashboard access to check the live domain, since the required env vars (GEMINI_API_KEY, Supabase service role key, etc.) aren't something I have locally.

Also , Since delete-account isn't referenced anywhere in the frontend, tests, or supabase/config.toml, I'd suggest removing it as part of resolving #555, so there's a single canonical deletion path going forward. Flagging in case it's intentionally kept for something outside this repo (an admin tool, ops script, etc.) that I wouldn't be able to see from here — let me know if that's the case and I'll leave it in place instead.

For any further updates, pls let me know, will definately work on them.

PR #604, opened 2026-07-14
feat: improve landing page navbar UI

priya434960:
## **Pull Request Description**

- Reduced the navbar height for a more compact appearance.
- Improved navigation spacing and alignment.
- Added a subtle active state indicator for navigation items.
- Updated the Sign In button styling to an outlined button.
- Refined the theme toggle container styling.
- Improved the alignment of authentication actions (Sign In / Get Started / Dashboard).

### **1. Type of Change**
- [ ] **Bug Fix** (non-breaking change which fixes an issue)
- [x] **New Feature** (non-breaking change which adds functionality)
- [ ] **Refactoring / Performance** (non-breaking code improvements)
- [ ] **Documentation Update** (changes to README, guides, or comments)
- [ ] **Other** (please specify): _________

### **2. Related Issue**
- Fixes #603 

### **3. How Has This Been Tested?**
Describe the tests/verification steps you ran to verify your changes.
- [x] **Local Build**: The application builds successfully locally (`npm run build`).
- [x] **Lint/Format Checks**: Local checks passed (`npm run lint` / `npm run format:check`).
- [x] **Manual Verification**: Briefly list the steps/features verified manually.
 

### **5. Contributor Quality Checklist**
- [x] My code follows the code style and formatting guidelines of this project.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no new warnings or console errors.
- [x] There is no commented-out code or debug logs left in the codebase.

PR #614, opened 2026-07-17
navbar improvement

priya434960:
Screenshot 2026-07-17 172322

Fixes #604


mohdmaazgani:
@priya434960 Thank you for the changes. However, the **"Features"** navbar item is highlighted by default on page load. Please remove the default active/highlight state and ensure that a navbar item is highlighted **only when the user hovers over it or when it is the currently selected/clicked section**. The initial state should have no item pre-highlighted unless the user is already on that section. Thank you!

PR #676, opened 2026-07-25
fix : use deterministic width in SidebarMenuSkeleton instead of Math.random()

### LINKED ISSUE #671: fix : use deterministic width in SidebarMenuSkeleton instead of Math.random()

Replace the `Math.random()` call in `src/components/ui/sidebar.tsx`'s `SidebarMenuSkeleton` component with a deterministic value. Using `Math.random()` in a React component causes a new random value on every render, which is wasteful and produces inconsistent skeleton animations.

**File:** `src/components/ui/sidebar.tsx` (SidebarMenuSkeleton component)

**Current code:**
```typescript
const width = React.useMemo(() => {
  return `${Math.floor(Math.random() * 40) + 50}%`;
}, []);
```

**Fix:** Replace the random computation with a deterministic value or a stable computed value. For example:
```typescript
const width = React.useMemo(() => {
  // Stable pseudo-random based on a fixed seed to vary skeleton widths without Math.random()
  const seed = 0x4d61766973; // "Mavis" in hex
  const result = ((seed * 1664525 + 1013904223) >>> 0) % 40 + 50;
  return `${result}%`;
}, []);
```
Or simply use a fixed string like `"65%"` for simplicity and consistency.

1. In `SidebarMenuSkeleton`, replace the `Math.random()`-based width computation with a deterministic alternative.
2. Ensure the change still produces a reasonable skeleton width (50-90% range).
3. Remove the `Math.random()` usage from this file.

- Consistent skeleton rendering across re-renders without triggering unnecessary re-computation.
- Eliminates a potential source of visual flicker.
- Removes reliance on `Math.random()` in a UI component where determinism is preferred.

tmdeveloper007:
## Summary of What Has Been Done

Replaced the `Math.random()` call in `src/components/ui/sidebar.tsx`'s `SidebarMenuSkeleton` component with a deterministic LCG using a fixed seed.

**File changed:** `src/components/ui/sidebar.tsx`

Closes #671


mohdmaazgani:
## Reviewer Suggestions @tmdeveloper007 

Overall, the PR is technically correct and addresses the issue of non-deterministic rendering caused by `Math.random()`. I have only a few minor, non-blocking suggestions:

1. **Consider simplifying the implementation**
   - Since the seed is fixed, the computed width will always be the same value.
   - Using a Linear Congruential Generator (LCG) adds complexity without providing additional benefits in this case.
   - A simple constant value (e.g. `"73%"`) or a predefined deterministic width would be easier to understand and maintain.

2. **Preserve visual variation while remaining deterministic**
   - Skeleton loaders generally look more natural when placeholder widths vary.
   - Consider using a small predefined array of widths (e.g. `["60%", "72%", "85%", "68%"]`) and selecting one deterministically based on the item index or another stable identifier.
   - This keeps the UI visually appealing while avoiding hydration issues.

3. **Add a brief explanation if the LCG approach is intentional**
   - If there is a specific reason for choosing an LCG instead of a simpler deterministic value, adding a short comment explaining the rationale would help future contributors understand the implementation.

---

## Question 

Thanks for the fix! I have one question regarding the implementation:

Since the seed is fixed and always produces the same output, was there a specific reason for using an LCG instead of a simpler deterministic approach (such as a constant width or a predefined deterministic width array)? I'm curious about the reasoning behind this design choice.

PR #848, opened 2026-08-01
Refactor origin validation and rate limiting logic

### LINKED ISSUE #807: [BUG] : broadcast-emergency lacks rate limiting and JWT verification, unlike every other Edge Function
## 🐛 Bug Description

Every other authenticated Edge Function in this project follows the same security pattern established in `delete-account`, `delete-user-account`, `get-cached-data`, `invalidate-cache`, and `symptom-analyzer`:
1. Check the `Origin` header against an allowlist (including the Netlify prod domain + preview regex).
2. Enforce a per-IP `rateLimit()` check via the shared `_shared/rateLimit.ts` helper.
3. Extract the bearer token and call `supabaseClient.auth.getUser(token)` to authenticate the caller, rejecting with `401` if it fails.

`supabase/functions/broadcast-emergency/index.ts` diverges from this pattern in three ways that compound into a real abuse vector, since this function's entire job is to spend money by sending an SMS through Twilio:

**1. No rate limiting at all.** There is no import of, or call to, `rateLimit()` anywhere in this file — compare to e.g. `delete-account/index.ts`:
```ts
const rateLimitResult = await rateLimit(ip);
if (!rateLimitResult.success) { ... return 429 ... }
```
`broadcast-emergency` has nothing equivalent. Any client that can obtain *any* valid Supabase-issued JWT (i.e., any signed-up user) can call this endpoint in a tight loop with a valid signature (self-signed with a freshly generated keypair each time — signature verification only checks that the payload matches the *provided* `publicKeyJwk`, it doesn't check that key against anything server-side) and trigger unlimited Twilio SMS sends, since `contact_phone` is taken directly from the request body.

**2. Authorization header presence is checked, but the JWT is never actually verified against Supabase.** The function does this:
```ts
const authHeader = req.headers.get("Authorization");
if (!authHeader) { ... return 401 ... }

const supabaseClient = createClient(SUPABASE_URL, SUPABASE_ANON_KEY, {
  global: { headers: { Authorization: authHeader } },
});

const token = authHeader.replace("Bearer ", "");
const { data: { user }, error: userError } = await supabaseClient.auth.getUser(token);

if (userError || !user) { ... return 401 ... }
```
This part actually *does* call `getUser()` — so on first read it looks fine. But note the `createClient` call passes `Authorization: authHeader` into `global.headers`, which (as correctly identified and commented on in `delete-user-account/index.ts`) does **not** populate an auth session — it only affects outgoing REST calls made *by that client instance*. `auth.getUser(token)` works here anyway because the `token` argument is passed explicitly, so functionally this call does succeed. The real issue is that this authentication check is currently pointless from a security standpoint given point #1: since there's no rate limiting, an authenticated-but-unlimited caller (any real user account, or one created for $0 via self-signup if signups are open) can hammer this endpoint.

**3. Missing the Netlify origin** (filed as part of a separate CORS issue), meaning the origin allowlist itself is already inconsistent with the rest of the codebase, suggesting this file was written before the shared security pattern was finalized and never back-filled.

## 🔄 Steps to Reproduce
1. **Sign up** for a normal user account (or use any existing session token).
2. **Generate** a fresh ECDSA P-256 keypair client-side, sign a well-formed emergency payload with your own attacker-controlled `contact_phone` (e.g. a premium-rate or attacker-owned number).
3. **Send** this signed payload to `broadcast-emergency` in a loop, e.g. 100 times in a few seconds.
4. **Expected Outcome:** after some small number of requests, the server responds `429 Rate limit exceeded`.
5. **Actual Outcome:** every request is processed and attempts a Twilio send (until Twilio's own account-level throughput limits kick in), because there is no per-IP or per-user throttling on this function at all.

## ✅ Expected Behavior
`broadcast-emergency` should be rate-limited the same way as every other Edge Function in this project, given that it triggers a real-world, billable side effect (an SMS send). A sensible limit here should probably be *stricter* than the generic `MAX_REQUESTS = 10`/minute used elsewhere — a legitimate user should essentially never need more than one or two emergency broadcasts per minute.

## 📌 System Information
- **Runtime:** Supabase Edge Functions (Deno)
- **Affected file:** `supabase/functions/broadcast-emergency/index.ts`

ishitaajain22-tech:
Fixes two related gaps in `broadcast-emergency` that made it diverge from the security pattern used by every other Edge Function in this project:

1. **No rate limiting** on an endpoint that triggers a billable Twilio SMS send, allowing unlimited-cost abuse from any authenticated user.
2. **Incomplete CORS allowlist** — missing the Netlify production origin and deploy-preview regex that every sibling function already includes, risking CORS-blocked calls from the deployed frontend.

Adds the shared `rateLimit()` helper (matching `delete-account`, `delete-user-account`, etc.) and aligns `ALLOWED_ORIGINS`/`isAllowedOrigin` with the rest of the codebase.

Fixes #807 

---

### **1. Type of Change**
- [ ] **New Feature**
- [ ] **Refactoring / Performance**
- [ ] **Documentation Update**
- [ ] **Other**: _________

---

- [x] **Local Build**: builds successfully (`npm run build`, `supabase functions serve` locally).
- [x] **Manual Verification**:
  1. Fired 15 rapid requests to `broadcast-emergency` from the same IP and confirmed a `429 Rate limit exceeded` response after the 10th.
  2. Confirmed requests from `https://symptom-scribe-clean.netlify.app` and deploy-preview URLs are now accepted (previously rejected with `403 Origin not allowed`).
  3. Confirmed legitimate, under-limit requests still succeed end-to-end (reach Twilio, return `200`).

**File changed:** `supabase/functions/broadcast-emergency/index.ts`

---

### DIFF
--- a/supabase/functions/broadcast-emergency/index.ts
+++ b/supabase/functions/broadcast-emergency/index.ts
@@ -1,15 +1,23 @@
 import { serve } from "https://deno.land/std@0.168.0/http/server.ts";
 import { createClient } from "https://esm.sh/@supabase/supabase-js@2.7.1";
+import { rateLimit } from "../_shared/rateLimit.ts";
 
 const ALLOWED_ORIGINS = [
   "http://localhost:3000",
   "http://localhost:8080",
   "https://symptom-scribe.vercel.app",
+  "https://symptom-scribe-clean.netlify.app",
 ];
+
+const NETLIFY_PREVIEW_ORIGIN = /^https:\/\/deploy-preview-\d+--symptom-scribe-clean\.netlify\.app$/;
+
+function isAllowedOrigin(origin: string): boolean {
+  return ALLOWED_ORIGINS.includes(origin) || NETLIFY_PREVIEW_ORIGIN.test(origin);
+}
 
 const getCorsHeaders = (origin: string | null) => ({
   "Access-Control-Allow-Origin":
-    origin && ALLOWED_ORIGINS.includes(origin) ? origin : "null",
+    origin && isAllowedOrigin(origin) ? origin : "null",
   "Access-Control-Allow-Headers":
     "authorization, x-client-info, apikey, content-type",
 });
@@ -22,7 +30,7 @@ function hexToUint8Array(hex: string): Uint8Array {
 serve(async (req) => {
   const origin = req.headers.get("origin");
 
-  if (origin && !ALLOWED_ORIGINS.includes(origin)) {
+  if (origin && !isAllowedOrigin(origin)) {
     return new Response(
       JSON.stringify({ error: "Origin not allowed" }),
       {
@@ -37,6 +45,20 @@ serve(async (req) => {
   }
 
   try {
+    // Rate limit before any expensive work — this endpoint triggers a billable Twilio send
+    const ip =
+      req.headers.get("x-forwarded-for") ||
+      req.headers.get("cf-connecting-ip") ||
+      "unknown";
+
+    const rateLimitResult = await rateLimit(ip);
+    if (!rateLimitResult.success) {
+      return new Response(
+        JSON.stringify({ error: "Rate limit exceeded. Please try again later." }),
+        { status: 429, headers: { ...getCorsHeaders(origin), "Content-Type": "application/json" } }
+      );
+    }
+
     const authHeader = req.headers.get("Authorization");
     if (!authHeader) {
       return new Response(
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
@ishitaajain22-tech Please resolve the merge conflicts.

ishitaajain22-tech:
@mohdmaazgani done

PR #859, opened 2026-08-01
Fix/dashboard initial load

### LINKED ISSUE #858: [BUG] Dashboard data not loading on initial login after auth redirect
## Summary
After logging into the application, the dashboard can occasionally fail to load user data on the initial load. A hard refresh (Ctrl+Shift+R) then loads the dashboard correctly.

## Expected behavior
After a successful sign-in, the dashboard should load the user's profile and history data without requiring a manual refresh.

## Actual behavior
On initial login, the dashboard can render before the authenticated session is fully available, causing the data fetch flow to skip or fail. Reloading the page rehydrates the app and the data appears correctly.

## Root cause
The dashboard initialization was not waiting for a validated authenticated session before fetching data. This created a race during the first login flow.

## Steps to reproduce
1. Sign out of the app.
2. Sign in with a valid account.
3. Navigate to the dashboard immediately after authentication.
4. Observe that the dashboard may load without the expected data.
5. Perform a hard refresh and the data appears correctly.

## 📌 System Information  

gssoc26

vib3withsimran:
A clear and concise description of the changes introduced by this pull request.

This PR improves the dashboard’s initial load experience by avoiding an unnecessary cache-first round trip for symptom history.
The dashboard now requests profile and symptom-history data in parallel and prefers the direct query path for faster first-load rendering.
This also preserves a fallback to cached data for empty or unavailable live results.

[right now loading time is around 2-3 seconds better than the previous one where we have to do ctrl+shift+r]

---

### **1. Type of Change**

---

### **2. Related Issue**
- Fixes #858 
---

- Added a regression test for the initial login/session hydration flow
- Verified the dashboard test suite passes
- Verified the production build succeeds

---

| Before | After |
| :--- | :--- |

---

### DIFF
--- a/src/lib/encryption.test.ts
+++ b/src/lib/encryption.test.ts
@@ -41,8 +41,20 @@ describe("Encryption Key Persistence & Security", () => {
     expect(getSearchKey()).toBe(key);
   });
 
-  it("generates P2P signing keys without persisting private key material to localStorage", async () => {
-    // Pre-populate legacy storage entries to verify cleanup
+  it("generates a P2P signing keypair when no key material exists", async () => {
+    const p2pKeys = await getP2PSigningKeys();
+
+    expect(p2pKeys.privateKey).toBeDefined();
+    expect(p2pKeys.publicKey).toBeDefined();
+
+    const privateJwk = await crypto.subtle.exportKey("jwk", p2pKeys.privateKey);
+    const publicJwk = await crypto.subtle.exportKey("jwk", p2pKeys.publicKey);
+
+    expect(privateJwk.kty).toBe("EC");
+    expect(publicJwk.kty).toBe("EC");
+  });
+
+  it("cleans up legacy P2P key entries and returns usable keys", async () => {
     localStorage.setItem("symptom_scribe_p2p_private_key", "fake-unencrypted-private-key");
     localStorage.setItem("symptom_scribe_p2p_enc_private_key", "fake-encrypted-private-key");
     localStorage.setItem("symptom_scribe_p2p_public_key", "fake-public-key");
@@ -51,10 +63,17 @@ describe("Encryption Key Persistence & Security", () => {
     expect(p2pKeys.privateKey).toBeDefined();
     expect(p2pKeys.publicKey).toBeDefined();
 
-    // Verify no private-key material is persisted to browser storage
     expect(localStorage.getItem("symptom_scribe_p2p_private_key")).toBeNull();
     expect(localStorage.getItem("symptom_scribe_p2p_enc_private_key")).toBeNull();
     expect(localStorage.getItem("symptom_scribe_p2p_public_key")).toBeNull();
+  });
+
+  it("reuses cached P2P signing keys on repeated calls", async () => {
+    const firstKeys = await getP2PSigningKeys();
+    const secondKeys = await getP2PSigningKeys();
+
+    expect(secondKeys.privateKey).toBe(firstKeys.privateKey);
+    expect(secondKeys.publicKey).toBe(firstKeys.publicKey);
   });
 
   it("signs and verifies emergency mesh payloads using P2P keypair", async () => {
--- a/src/lib/encryption.test.ts
+++ b/src/lib/encryption.test.ts
@@ -7,11 +7,15 @@ import {
   encryptText,
   decryptText,
   deriveKeyFromToken,
+  getP2PSigningKeys,
+  signPayload,
+  verifyPayload,
 } from "./encryption";
 
-describe("Encryption Key Persistence", () => {
+describe("Encryption Key Persistence & Security", () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    localStorage.clear();
   });
 
   it("successfully encrypts and decrypts text using derived keys", async () => {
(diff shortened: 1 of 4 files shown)


mohdmaazgani:
Hi @vib3withsimran, thank you for your contribution. I noticed that this and the previous PR(s) include changes related to multiple issues. To keep the review process clear and manageable, I'd kindly request that you raise separate PRs for each individual issue. This helps with easier review, testing, and tracking of changes. Thank you for your understanding!

vib3withsimran:
hi @mohdmaazgani 
I guess you are talking about multiple commits in a single PR. 
Yes, There's some issue from my side currently in pushing the commits. I tried to resolve it, but it didn't work. So instead, I tried to give clearer names to each commit so it would be easier to review. 
I would try to resolve it for the next upcoming PRs.

mohdmaazgani:
Thank you for your contribution! The overall direction looks good, and the changes improve both security and performance. Before this PR is ready to merge, I have a few suggestions that I believe should be addressed.

## 1. Clarify intent for public key storage

The removal of legacy private keys from `localStorage` is a good security improvement. However, the change also removes the persisted public key.

Since public keys are not sensitive, removing them changes the application's behavior by creating a new public identity after every page refresh.

Could you clarify whether this was intentional?

If so, please:

- Add a short explanation in the PR description describing the migration.
- Add a brief code comment explaining that public identity will no longer persist across browser reloads.

If this was unintentional, consider restoring persistence for the public key while continuing to keep the private key in memory only.

---

## 2. Consider generating non-extractable private keys (Security Hardening)

The current implementation is valid and works correctly. However, you could further strengthen security by generating the private key as **non-extractable** unless exporting it is actually required.

```ts
crypto.subtle.generateKey(
  {
    name: "ECDSA",
    namedCurve: "P-256",
  },
  false,
  ["sign", "verify"]
);
```

This would prevent accidental future exports of the private key.

> **Note:** The current tests export the generated keys, so they would need to be updated if this change is adopted.

---

## 3. Prevent duplicate concurrent key generation

There is a small race condition where multiple concurrent callers could observe that the cache is empty and generate separate keypairs before one is cached.

Consider introducing an in-flight promise (for example, `cachedP2PKeysPromise`) so that:

- The first caller generates the keypair.
- Subsequent callers await the same promise.
- Duplicate work and inconsistent state are avoided.

This is not a blocker, but it would make the implementation more robust.

---

## 4. Improve robustness around `localStorage` cleanup

The existing `typeof localStorage !== "undefined"` guard already prevents SSR issues.

As an additional robustness improvement, consider wrapping the legacy `removeItem()` calls in a `try/catch` block since storage operations may throw in some restricted browser environments.

Any cleanup failure should be logged but should not prevent key generation.

Example:

```ts
if (typeof localStorage !== "undefined") {
  try {
    localStorage.removeItem("symptom_scribe_p2p_private_key");
    localStorage.removeItem("symptom_scribe_p2p_enc_private_key");
    localStorage.removeItem("symptom_scribe_p2p_public_key");
  } catch (err) {
    console.warn("Failed to remove legacy P2P keys:", err);
  }
}
```

---

## 5. Ensure the loading state is always cleared

`fetchDashboardData()` calls `setLoading(true)` before performing several asynchronous operations.

To ensure the loading indicator is never left active after an unexpected error, consider wrapping the function body in a `try/finally` block:

```ts
setLoading(true);

try {
  // fetch and decrypt data
} finally {
  setLoading(false);
}
```

This guarantees the loading state is reset even if a network request or decryption step throws unexpectedly.

---

## 6. Avoid redundant session lookups

`initializeDashboardSession()` already resolves the authenticated session before calling `fetchDashboardData()`.

Since `fetchDashboardData()` already accepts an optional user ID, consider using the provided value directly instead of performing another `getSession()` call when it is already available.

This removes an unnecessary authentication lookup, simplifies the control flow, and avoids an extra asynchronous call.

---

Overall, this is a solid implementation. Addressing the points above would further improve the security, robustness, performance, and maintainability of the changes while keeping the intended behavior well documented.

vib3withsimran:
Thanks for the review. @mohdmaazgani 
I intentionally changed the P2P key flow to be session-scoped, so the public identity no longer persists across browser reloads. I’ve added a short code comment to make that behaviour explicit and will note the migration in the PR description.

I also hardened the key generation path by making the private key non-extractable, added protection against duplicate concurrent key generation with an in-flight promise, wrapped legacy localStorage cleanup in try/catch, and removed the redundant session lookup in the dashboard loader. The dashboard loading state is now cleared in a finally block as well.

I updated the regression tests to match the non-extractable private key behaviour and the concurrent key-generation guard.

I verified the changes with the encryption tests, dashboard tests, and a production build.

Just a small request- once you verify that the entire codebase is correct, can you just merge it and add the required tags as soon as possible? As my GSSOC streak is at stake, as I was not available for the last complete week.

mohdmaazgani:
@vib3withsimran Okay....

> Thanks for the review. @mohdmaazgani I intentionally changed the P2P key flow to be session-scoped, so the public identity no longer persists across browser reloads. I’ve added a short code comment to make that behaviour explicit and will note the migration in the PR description.
> 
> I also hardened the key generation path by making the private key non-extractable, added protection against duplicate concurrent key generation with an in-flight promise, wrapped legacy localStorage cleanup in try/catch, and removed the redundant session lookup in the dashboard loader. The dashboard loading state is now cleared in a finally block as well.
> 
> I updated the regression tests to match the non-extractable private key behaviour and the concurrent key-generation guard.
> 
> I verified the changes with the encryption tests, dashboard tests, and a production build.
> 
> Just a small request- once you verify that the entire codebase is correct, can you just merge it and add the required tags as soon as possible? As my GSSOC streak is at stake, as I was not available for the last complete week.

mohdmaazgani:
Hi @vib3withsimran! I’ve merged your PR. 🎉

Just wanted to check, has your GSSoC streak been restored?

Also, if something like this happens again, just let me know beforehand. You can always raise a small README/documentation PR if you're unable to contribute code for a while. Since you've been associated with the project for quite some time, I can review and merge it quickly so you don't lose your streak unnecessarily.

Feel free to reach out anytime if you're in a similar situation. Happy to help. 😊

mohdmaazgani:
Hi @vib3withsimran ! Just wanted to let you know that after merging your PR, two contributors raised concerns—one about some unnecessary file changes and another about the `encryption.test.ts` test failures. To be on the safe side and verify that everything was working fine before the merge, I temporarily reverted the PR.

I also wanted to ask—were you seeing the same issues on your end? Did you also notice the encryption.ts test failures or the additional file changes while working on the PR, or was everything passing normally for you? I'd appreciate your thoughts so we can figure out what's going on together. Thanks! 😊

image

Screenshot 2026-08-03 132849

vib3withsimran:
Hi @mohdmaazgani 

I guess my streak has been broken. I was supposed to get done with at least one PR on Saturday. But I started working on the issues on Saturday and get merged by mid-Sunday/Monday. 

I will keep your suggestion in mind. Thank you for your concern

I will also look into the issue related to `encryption.test.ts' if something is still problematic in there.

mohdmaazgani:
@vib3withsimran I'm really sorry if this happened because of my late merging or from my end.

mohdmaazgani:
@vib3withsimran Please firstly solve the issue of the dashboard loading one.....

vib3withsimran:
No no @mohdmaazgani 
it's not a problem 

GSSOC leaderboard is also not updated right now. So it was a mild guess from my side. I am also not able to check it.

Is dashboard loading causing still error? I have tried to solve it again. This time dashboard was taking like 2-3 second to load, and not required to do 'ctrl + shift + r'. Can you check it once and confirm me if any other improvement needed for the dashboard loading. If yes, I will give it one more try then.

mohdmaazgani:
@vib3withsimran Since I had to revert your PR, all of those changes have been reverted as well.
So please check it again...yes, it's still loading since minutes.
image

vib3withsimran:
hi @mohdmaazgani 

I noticed the dashboard has completely changed. Is this done under ur suivilience or someone has made it randomly? can u check this and revert to whoever PR it was? creating complete new dahsboard ui interface would not be a good choice. 

image

mohdmaazgani:
hi @vib3withsimran , I guess the live project is working fine.
Are you on the latest commit ?
Or please tell me the path where it is actually occurring.
image

mohdmaazgani:
@vib3withsimran Is everything fine now.....or the above issue is still there ? Please let me know....

vib3withsimran:
@mohdmaazgani 
Yeah, now everything is fine.
There was something wrong on my side. Thank you for clearing that. 
I have to fetch all commits and then reset hard the branch.

mohdmaazgani:
Okay @vib3withsimran....thanks for the clarification.

PR #906, opened 2026-08-03
feat(#776): debounce and rate limit symptom analysis API calls

### LINKED ISSUE #776: [Feature]: Implement rate limiting and request debouncing on symptom analysis API calls to prevent abuse and reduce unnecessary Supabase usage
## Summary

The symptom analysis and AI insight features in Symptom Scribe make API calls to Supabase on user interactions. Currently there is no client-side debouncing on these calls and no rate limiting enforced, meaning a user — or a script — can trigger repeated expensive requests in rapid succession. This creates unnecessary Supabase read/write load and increases operational cost.

## Problem

- Symptom logging and analysis calls fire immediately on user interaction with no debounce delay
- There is no client-side guard preventing the same request from being sent multiple times in quick succession (e.g., double-clicking a submit button)
- No per-user rate limit is applied at the API layer, making the app susceptible to accidental or intentional call flooding
- Supabase free-tier row limits and API call budgets can be exhausted quickly under this pattern

## Impact

- Unnecessary Supabase API quota consumption and potential overage billing
- Degraded UX if the Supabase backend starts rate-limiting responses
- Risk of duplicate symptom entries being inserted if a user submits multiple times before a response arrives

## Proposed Solution

I will implement the following across the relevant hooks and components:

**1. Client-side debouncing using a custom `useDebounce` hook**
```ts
// src/hooks/useDebounce.ts
import { useEffect, useState } from 'react';

export function useDebounce<T>(value: T, delay: number): T {
  const [debouncedValue, setDebouncedValue] = useState(value);
  useEffect(() => {
    const timer = setTimeout(() => setDebouncedValue(value), delay);
    return () => clearTimeout(timer);
  }, [value, delay]);
  return debouncedValue;
}
```

**2. Submission guard using a `useRef` flag** to prevent duplicate submissions while a request is in-flight:
```ts
const isSubmitting = useRef(false);

const handleSubmit = async () => {
  if (isSubmitting.current) return;
  isSubmitting.current = true;
  try {
    await submitSymptomData(data);
  } finally {
    isSubmitting.current = false;
  }
};
```

**3. Optional: Supabase edge function rate limiting** using a simple per-user call count stored in a dedicated table, checked before processing AI requests.

I will apply these patterns to all symptom submission and AI analysis interaction points in `src/components/` and `src/pages/`.

**Labels:** `enhancement`, `performance`, `good first issue`, `GSSoC 2026`

Would you assign this issue to me? Happy to implement end-to-end.

Anijesh:

Symptom analysis calls the `symptom-analyzer` Supabase edge function on every user interaction. There was no debounce and no client-side guard, so a double-click, a stuck Enter key, or a script could fire several identical analyses back to back — burning Supabase/AI quota and risking duplicate rows in `symptom_history`.

This PR adds debouncing and a submission guard on the client, and a per-user bucket to the rate limiter on the API side.

### What changed

**1. `useDebounce` hook (`src/hooks/useDebounce.ts`)**
Generic hook that publishes a value only once it has stayed unchanged for the given delay, so a burst of updates collapses into a single downstream effect.

**2. `useSubmitGuard` hook (`src/hooks/useSubmitGuard.ts`)**
Wraps a submission in three checks before it is allowed to run:
- **In-flight guard** — a `useRef` flag rejects a second submission while the first is still awaiting a response. A `useState` loading flag cannot do this on its own, because it only lands on the next render, which leaves a double-click window open.
- **Cooldown** — a minimum gap (1.5s) between two accepted submissions.
- **Rolling quota** — at most 8 accepted submissions per minute, matching the budget the edge function enforces.

Rejections never throw: the caller reads `outcome.status` and shows the message from `getRejectionMessage()`.

**3. Guarded the symptom analysis entry points**
- `src/components/chat/ChatInterface.tsx` — `handleSend`
- `src/pages/Health/AIHealthAssistant.tsx` — `handleAnalyze` (also covers the suggestion chips, which are easy to click repeatedly)

A blocked request now shows a "Too many requests" warning toast instead of silently doing nothing.

**4. Debounced the dashboard predictions (`src/components/dashboard/SymptomPredictions.tsx`)**
This call fires from an effect whenever the symptom list changes. It is now debounced by 800ms and keyed on a hash of the list, so a new array holding the same symptoms no longer refires it. Duplicate in-flight requests are dropped, and a response for a symptom list that has since moved on is discarded instead of overwriting fresher predictions.

**5. Per-user rate limiting (`supabase/functions/symptom-analyzer/index.ts`)**
The function only bucketed by IP, so one account could spread a flood across devices, tabs, or networks, while users behind a shared IP (offices, mobile carriers, NAT) were throttled collectively. It now checks an IP bucket **and** a per-authenticated-user bucket, and returns `Retry-After: 60` alongside the 429.

**6. 429 handling in the UI**
All three call sites now recognise a 429 and surface a clear "Too many requests" message instead of a generic failure.

---

### **1. Type of Change**
- [ ] **Bug Fix** (non-breaking change which fixes an issue)
- [x] **New Feature** (non-breaking change which adds functionality)

---

### **2. Related Issue**
Fixes #776

---

- [x] **Local Build**: `npm run build` succeeds.
- [x] **Lint/Format Checks**: `npm run lint` reports 0 errors (only the 9 pre-existing `react-refresh` warnings in files this PR does not touch). New files pass `prettier --check`.
- [x] **Unit Tests**: `npm test` — **91 passed**, including **12 new tests** for the two hooks (`src/hooks/useDebounce.test.ts`, `src/hooks/useSubmitGuard.test.ts`) covering the debounce window, burst collapsing, timer cleanup, the in-flight guard, the cooldown, the rolling quota and its recovery, and guard release when a task throws.
- [x] **Type-check**: no new TypeScript errors — the error list is byte-identical to `main` apart from two pre-existing errors whose line numbers shifted.
- [x] **Manual Verification**:
  1. AI chat — clicked **Send** rapidly / held Enter: only one request leaves the browser, the extra attempts show the warning toast, and only one entry is written to symptom history.
  2. AI Health Assistant — clicked several suggestion chips in quick succession: same behaviour.
  3. Dashboard — the predictions call now fires once after the symptom list settles rather than on every intermediate update.

> **Note on `useSubmitGuard`'s result type:** it uses a string discriminant (`status: "accepted" | "rejected"`) rather than a boolean flag, because this project compiles with `strictNullChecks: false`, where TypeScript cannot narrow a union on a boolean literal.

---

No layout changes. The only visible addition is a warning toast when a request is throttled:

> **Too many requests** — Your previous request is still being analyzed. Please wait for it to finish.

---

### DIFF
--- a/src/components/chat/ChatInterface.tsx
+++ b/src/components/chat/ChatInterface.tsx
@@ -32,6 +32,7 @@ import {
   shouldPersistConsultation,
 } from "@/lib/symptom-consultation";
 import { compressImage, uploadSymptomImage } from "@/lib/image-utils";
+import { useSubmitGuard, getRejectionMessage } from "@/hooks/useSubmitGuard";
 import ChatLoading from "./ChatLoading";
 import { Sheet, SheetContent, SheetHeader, SheetTitle, SheetTrigger } from "@/components/ui/sheet";
 import { type Json } from "@/integrations/supabase/types";
@@ -82,6 +83,10 @@ const ChatInterface = () => {
   const messagesEndRef = useRef<HTMLDivElement>(null);
   const fileInputRef = useRef<HTMLInputElement>(null);
   const { toast } = useToast();
+
+  // Throttles the analysis endpoint: no overlapping requests, a short cooldown
+  // between sends, and a per-minute budget.
+  const runGuardedSend = useSubmitGuard();
 
   const handleFileChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     if (e.target.files) {
@@ -168,9 +173,7 @@ const ChatInterface = () => {
     setIsMobileOpen(false);
   };
 
-  const handleSend = async () => {
-    if (!input.trim() || isLoading) return;
-
+  const sendMessage = async () => {
     const userMessage: Message = { role: "user", content: input.trim() };
     const newMessages = [...messages, userMessage];
     setMessages(newMessages);
@@ -251,6 +254,10 @@ const ChatInterface = () => {
         },
         body: JSON.stringify({ messages: [...recentContext, userMessage] }),
       });
+
+      if (response.status === 429) {
+        throw new Error("Too many requests. Please wait a moment before trying again.");
+      }
 
       if (!response.ok || !response.body) {
         throw new Error("Failed to start stream");
@@ -426,6 +433,16 @@ const ChatInterface = () => {
       setIsLoading(false);
       setSelectedFiles([]);
       if (fileInputRef.current) fileInputRef.current.value = "";
+    }
+  };
+
+  const handleSend = async () => {
+    if (!input.trim() || isLoading) return;
+
+    const outcome = await runGuardedSend(sendMessage);
+
+    if (outcome.status === "rejected") {
+      showWarning("Too many requests", getRejectionMessage(outcome.reason, outcome.retryAfterMs));
     }
   };
 
--- a/src/components/dashboard/SymptomPredictions.tsx
+++ b/src/components/dashboard/SymptomPredictions.tsx
@@ -1,9 +1,10 @@
-import { useState, useEffect } from "react";
+import { useState, useEffect, useMemo, useRef } from "react";
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
 import { Badge } from "@/components/ui/badge";
 import { Sparkles, AlertTriangle, CheckCircle, Info, Loader2 } from "lucide-react";
 import { supabase } from "@/integrations/supabase/client";
 import { browserEnv } from "@/lib/env";
+import { useDebounce } from "@/hooks/useDebounce";
 
 interface Prediction {
   risk: string;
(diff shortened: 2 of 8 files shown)


Anijesh:
Merge conflict resolved — rebased onto the latest `main` (`0c8bb86`). The branch is now mergeable.

Two files conflicted; both were resolved by keeping the incoming changes from `main` alongside this PR's work, not by overwriting either side:

**`src/components/chat/ChatInterface.tsx`**
- The image-attachment work from #961 landed on the same lines as this PR. Kept both the `compressImage`/`uploadSymptomImage` import and `handleFileChange`/`removeFile`, next to the `useSubmitGuard` import and hook call.
- `handleSend` still wraps the full send path — including the new image upload — in the submit guard, so an attachment upload cannot be fired twice by a double-click either.

**`src/components/dashboard/SymptomPredictions.tsx`**
- `main` had gained a more detailed error-message chain. Rather than short-circuiting it with a separate 429 throw, the 429 case is now a branch inside that chain, so all of the existing 401/403, network, and 5xx messages are preserved.
- The `TypeError`/network branch from `main` is kept in the `catch`, now wrapped in the `isStale()` check so a response for a symptom list that has since moved on no longer overwrites fresher state.

Verification after the rebase:
- `npm run build` — succeeds
- `npm test` — **176 passed** (12 files), including the 12 tests for the two new hooks
- `npm run lint` — **0 errors**; the 13 remaining warnings are all pre-existing and none are in files this PR touches
- `npx tsc --noEmit` — error list is identical to `main` (42 on both), so no new type errors

All three CI checks are green.

@mohdmaazgani could you add the `GSSoC` label when you get a chance? I don't have permission to add it myself, and it's needed for the contribution to be counted.

PR #928, opened 2026-08-04
fix(metrics-form): persist logged metrics to Supabase

webgo-oss:
Problem
`src/components/forms/MetricsForm.tsx` had a `// TODO: Add your actual Supabase insert logic here` stub. The form validated correctly and showed `alert("Metrics logged successfully!")`, but the actual insert was commented out — so submissions were silently discarded while the user was told it worked.

Fix
- Wired the form to `supabase.from("health_metrics").insert(...)`, writing one row per metric (`steps`, `hydration`, `calories`) with `metric_type` + `value` JSON, matching the existing `health_metrics` table schema used elsewhere in the app (e.g. `src/pages/Metrics/index.tsx`).
- Added an auth check (`supabase.auth.getUser()`) so unauthenticated submissions fail gracefully with a toast instead of an opaque Supabase error.
- Replaced `alert()` with the app's `showSuccess` / `showError` toast helpers for consistent UX.
- Replaced raw HTML `input`/`button` with the shadcn `Input`/`Label`/`Button` components so the form matches the rest of the app's styling and works correctly in dark mode.
- Form now resets after a successful submit.

Scope
Only `src/components/forms/MetricsForm.tsx` is touched in this PR. The two duplicate/dead copies (`src/app/dashboard/components/HealthForm.tsx` and `src/pages/dashboard/index.tsx`) are unused (not referenced from `App.tsx` or anywhere else) and will be removed in a follow-up PR.

Testing
- Verified no syntax/build errors via esbuild parse check.
- Manual walkthrough of the submit path: unauthenticated → error toast, no insert attempted; authenticated with valid input → 3 rows inserted into `health_metrics`, success toast, form resets; Supabase insert failure → error toast with message, form retains entered values.


MathivadhaniArul:
/claim
Proposed Solution

Yet to fill

PR #1006, opened 2026-08-06
fix: add missing images column to symptom_history (#996)

### LINKED ISSUE #996: [BUG] Chat with attached image fails to save to symptom_history -- table has no images column
﻿`ChatInterface` builds a record containing `images: uploadedImages` and inserts it into `symptom_history`, but the table has no `images` column (it is missing from both the generated Supabase types and every migration). Every consultation with an attached image fails to save server-side, and the offline fallback can never sync.

Location:
- `src/components/chat/ChatInterface.tsx:359-371` ? record built with `images: uploadedImages` (line 370)
- `src/components/chat/ChatInterface.tsx:384-386` ? `supabase.from("symptom_history").insert(supabaseRecord)`
- `src/lib/offline-db.ts:84` ? `encryptSymptom` spreads the record, so `images` survives into the insert payload
- `src/lib/offline-db.ts:288-297` ? `syncOfflineData` retries the same insert (still containing `images`) on every session, so the record can never sync
- `src/integrations/supabase/types.ts:128-157` ? `symptom_history` Row/Insert has no `images` field
- No SQL migration under `supabase/migrations/` defines an `images` column (verified by searching all migrations)

Impact:
1. Online: the insert fails (Postgres 42703 / PGRST204 "Could not find the images column").
2. The failure path (`ChatInterface.tsx:388-402`) saves the record to Dexie with `pending_sync: 1`, and `syncOfflineData` retries the identical broken insert forever ? the record never reaches the server and never appears on other devices.

Suggested fix:
- Add a nullable `images text[]` column to `symptom_history` via a migration, or drop `images` from the insert payload (and the image upload) until the schema supports it.

ionfwsrijan:
## Problem

The symptom submission flow builds a `symptom_history` record that includes an `images` array, but the `symptom_history` table has no `images` column and the generated Supabase types (`src/integrations/supabase/types.ts`) do not include it. Inserts therefore fail because the payload references a column that does not exist in the schema or the types.

## Fix

- Adds a migration `supabase/migrations/20260806000000_add_images_to_symptom_history.sql` that adds a nullable `images TEXT[]` column to `symptom_history`.
- Adds `images` to the generated `symptom_history` types (`Row`, `Insert`, `Update`) in `src/integrations/supabase/types.ts`.
- Adds `images?: string[] | null` to the `OfflineSymptom` type in `src/lib/offline-db.ts` so offline records that carry images are typed correctly.

## Files changed

- `src/integrations/supabase/types.ts`
- `src/lib/offline-db.ts`
- `supabase/migrations/20260806000000_add_images_to_symptom_history.sql`

## Testing

- Applied the migration (nullable column, backward compatible).
- Manual QA: submit a symptom entry with images attached and verify the row is written to `symptom_history` including the `images` array; verify entries without images still insert correctly.

Closes #996

### DIFF
--- a/src/integrations/supabase/types.ts
+++ b/src/integrations/supabase/types.ts
@@ -131,6 +131,7 @@ export type Database = {
           created_at: string | null
           follow_up_date: string | null
           id: string
+          images: string[] | null
           possible_causes: string[] | null
           recommendations: string[] | null
           resolved: boolean | null
@@ -145,6 +146,7 @@ export type Database = {
           created_at?: string | null
           follow_up_date?: string | null
           id?: string
+          images?: string[] | null
           possible_causes?: string[] | null
           recommendations?: string[] | null
           resolved?: boolean | null
@@ -159,6 +161,7 @@ export type Database = {
           created_at?: string | null
           follow_up_date?: string | null
           id?: string
+          images?: string[] | null
           possible_causes?: string[] | null
           recommendations?: string[] | null
           resolved?: boolean | null
--- a/src/lib/offline-db.ts
+++ b/src/lib/offline-db.ts
@@ -38,6 +38,7 @@ export interface OfflineSymptom {
   pending_delete: number;
   ai_analysis?: string;
   search_tokens?: string[] | null;
+  images?: string[] | null;
 }
 
 export interface MeshAlert {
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

PR #1009, opened 2026-08-06
fix: re-encrypt server-side data after password change/reset (#999)

### LINKED ISSUE #999: [BUG] Changing/resetting the password never re-encrypts server-side data -- existing records become undecryptable
﻿After a password change, new encryption keys are derived from the new password, but only the local Dexie cache is re-encrypted. All server rows (`symptom_history`, `health_metrics`, encrypted `profiles` fields) remain encrypted under the old key and can no longer be decrypted after a reload.

Location:
- `src/pages/User/Settings.tsx:104-129` ? `updateUser({password})` then `setupKeysFromPassword(newPassword, ...)` then `triggerKeyRotation(oldKey, newKey, ...)`
- `src/lib/offline-db.ts:191-205` ? the `onTokenRefresh` hook re-encrypts ONLY `db.healthMetrics` and `db.symptomHistory` (local tables)
- `src/pages/User/ResetPassword.tsx:28-40` ? the reset page calls `updateUser({ password })` and never calls `setupKeysFromPassword`, so the in-memory key is not even rotated
- No code path re-encrypts server-side rows after a key change (no edge function, no client routine)

Impact: after changing or resetting the password, reloading the app derives the new key and AES-GCM decryption of previously stored server data fails, so History, Metrics, and Profile show nothing or "Connection Error". The `ResetPassword` flow additionally leaves the old key active in the current session.

Suggested fix: re-encrypt server rows during key rotation (lazily or via a migration), or add key-versioning to support migration; at minimum `ResetPassword` should call `setupKeysFromPassword`.

ionfwsrijan:
## Problem

After a password change or reset, only the local Dexie cache is re-encrypted (via the `onTokenRefresh` hook). The server-side rows — `symptom_history`, `health_metrics`, and the encrypted `profiles` fields — stay encrypted under the old key. After a reload the old key no longer exists, so previously saved records become undecryptable (broken profile, history, metrics, and server-side search).

## Fix

- Adds `reencryptServerData(oldKey, newKey, oldSearchKey, newSearchKey)` in `src/lib/offline-db.ts` which decrypts and re-encrypts all of the authenticated user's `symptom_history`, `health_metrics`, and encrypted `profiles` fields (text + array fields) with the new key and new search tokens, then invalidates the related caches.
- Calls `reencryptServerData` after `triggerKeyRotation` in the password change flow (`src/pages/User/Settings.tsx`).
- Calls `reencryptServerData` in the reset-password flow (`src/pages/User/ResetPassword.tsx`) after `setupKeysFromPassword` derives the new key, so records created under the old password are migrated too.
- Adds the missing `search_tokens` column to the generated types for `symptom_history` and `health_metrics` (the column already exists in the database per `20260616143000_add_search_tokens_blind_indexes.sql`), which the re-encryption writes.

## Files changed

- `src/integrations/supabase/types.ts`
- `src/lib/offline-db.ts`
- `src/pages/User/Settings.tsx`
- `src/pages/User/ResetPassword.tsx`

## Testing

- Manual QA: change the password, reload the app, and verify history/metrics/profile still decrypt correctly and search still returns rows; repeat for the reset-password flow.

Closes #999

### DIFF
--- a/src/integrations/supabase/types.ts
+++ b/src/integrations/supabase/types.ts
@@ -48,6 +48,7 @@ export type Database = {
           metric_type: string
           notes: string | null
           recorded_at: string | null
+          search_tokens: string[] | null
           user_id: string
           value: Json
         }
@@ -57,6 +58,7 @@ export type Database = {
           metric_type: string
           notes?: string | null
           recorded_at?: string | null
+          search_tokens?: string[] | null
           user_id: string
           value: Json
         }
@@ -66,6 +68,7 @@ export type Database = {
           metric_type?: string
           notes?: string | null
           recorded_at?: string | null
+          search_tokens?: string[] | null
           user_id?: string
           value?: Json
         }
@@ -135,6 +138,7 @@ export type Database = {
           recommendations: string[] | null
           resolved: boolean | null
           risk_score: number | null
+          search_tokens: string[] | null
           severity_level: string
           symptoms: string
           updated_at: string | null
@@ -149,6 +153,7 @@ export type Database = {
           recommendations?: string[] | null
           resolved?: boolean | null
           risk_score?: number | null
+          search_tokens?: string[] | null
           severity_level: string
           symptoms: string
           updated_at?: string | null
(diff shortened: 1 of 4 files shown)


ionfwsrijan:

PR #1010, opened 2026-08-06
fix: stop silently overwriting local salt on session change (#1000)

### LINKED ISSUE #1000: [BUG] handleSessionChange silently overwrites the local PBKDF2 salt when it differs from the DB -- previously encrypted data becomes unreadable
﻿`handleSessionChange` syncs the DB `encryption_salt` into localStorage and, whenever `dbSalt !== localSalt`, unconditionally overwrites the local salt before re-deriving the key. The code's own design comment states that changing the salt "will silently produce a different key and make all previously encrypted fields and blind-index search tokens unreadable" ? which is exactly what this overwrite does. No re-encryption is performed.

Location:
- `src/lib/encryption.ts:400-403` ? `if (dbSalt !== localSalt) localStorage.setItem(SALT_KEY_PREFIX + userId, dbSalt)`
- `src/lib/encryption.ts:447-456` ? key re-derived from the (possibly overwritten) salt immediately after
- `src/lib/encryption.ts:97-101` ? comment documenting that salt rotation breaks all existing data
- `src/lib/encryption.ts:105-118` ? `getUserSalt` reads the local salt used by `deriveKeyFromToken` (line 140)

Trigger: the DB salt diverges from a device's local salt when the salt upsert fails on one device (the `Promise.all(...).catch` at lines 423-425 swallows the error) or when two devices race while the DB salt is null (last-write-wins) ? exactly the cross-device scenario this code claims to support.

Impact: on the affected device the key silently changes and all locally cached and server-stored encrypted data becomes permanently undecryptable, with no migration path.

Suggested fix: treat a salt mismatch as an error requiring explicit re-encryption instead of silently swapping keys.

ionfwsrijan:
## Problem

`handleSessionChange` blindly overwrites the local salt with the database salt whenever the two differ. Because the salt is mixed into PBKDF2 key derivation, overwriting the local salt silently changes the derived key: data encrypted with the old (local) salt can no longer be decrypted after a reload, and existing offline data becomes unrecoverable.

## Fix

- Refuses to overwrite the local salt when `dbSalt !== localSalt`; instead it logs a clear error explaining that the salts differ and that re-encryption is required, so the user's existing data stays decryptable.
- Only adopts the database salt when no local salt exists (fresh device/first session), preserving the intended cross-device salt convergence for new users.

## Files changed

- `src/lib/encryption.ts`

## Testing

- Manual QA: on an existing device with local salt, trigger a session refresh and confirm the local salt is not replaced; on a brand-new device, confirm the database salt is adopted.

Closes #1000

### DIFF
--- a/src/lib/encryption.ts
+++ b/src/lib/encryption.ts
@@ -398,7 +398,22 @@ async function handleSessionChange(session: Session) {
     const localSalt = localStorage.getItem(SALT_KEY_PREFIX + userId);
 
     if (dbSalt) {
-      if (dbSalt !== localSalt) {
+      if (localSalt && dbSalt !== localSalt) {
+        // A diverged salt must not be silently swapped in: the salt is a
+        // direct input to key derivation, so overwriting it would silently
+        // produce a different key and make all previously encrypted fields
+        // and blind-index search tokens unreadable (see the "Per-user PBKDF2
+        // Salt" note above). Treat the mismatch as an error that requires an
+        // explicit re-encryption instead of rotating the key under the user.
+        console.error(
+          "Encryption salt mismatch: local salt differs from the database salt for this user. " +
+            "Refusing to overwrite the local salt because it would change the derived key and " +
+            "make existing encrypted data undecryptable. An explicit re-encryption is required " +
+            "to migrate to the database salt."
+        );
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #1011, opened 2026-08-06
fix: enforce AAL2 on protected routes after reload (#1001)

### LINKED ISSUE #1001: [BUG] ProtectedRoute does not enforce MFA/AAL2 -- reloading the app bypasses two-factor authentication
﻿MFA is only enforced in the sign-in flow on the Auth page. `ProtectedRoute` checks only that a session exists, so a user with 2FA enabled can reload the app or open a new tab and reach /dashboard, /chat, /history, etc. with an AAL1 session and no second-factor challenge.

Location:
- `src/components/auth/ProtectedRoute.tsx:20-33` ? only calls `supabase.auth.getSession()`
- `src/components/auth/ProtectedRoute.tsx:73-77` ? grants access if any session exists
- `src/pages/Auth/index.tsx:128-173` ? the only place AAL2/MFA is checked (challenge + verify during sign-in)

Impact: 2FA provides no protection once the session is restored from a refresh or a new tab.

Suggested fix: check `supabase.auth.mfa.getAuthenticatorAssuranceLevel()` in `ProtectedRoute` and require AAL2 when the account has a verified factor (redirect to an MFA challenge route).

ionfwsrijan:
## Problem

`ProtectedRoute` only checks that a session exists. After signing in with password only (AAL1), or right after the MFA challenge on the Auth page, a page reload passes `ProtectedRoute` with an AAL1 session, so protected routes can be reached without completing the 2FA (AAL2) step.

## Fix

- `ProtectedRoute` now calls `supabase.auth.mfa.getAuthenticatorAssuranceLevel()` and, when `nextLevel` is `aal2` but the current level is `aal1`, redirects to `/auth?mfa=challenge` instead of rendering the protected route.
- `src/pages/Auth/index.tsx` handles the `mfa=challenge` query parameter: it lists the user's TOTP factors and shows the MFA verification UI, and only derives encryption keys in `handleMfaVerify` when a password is available (so challenge-only sessions don't break key setup).

## Files changed

- `src/components/auth/ProtectedRoute.tsx`
- `src/pages/Auth/index.tsx`

## Testing

- Manual QA: sign in with password only, reload a protected page, and confirm redirection to `/auth?mfa=challenge`; complete the TOTP code and confirm access is granted.

Closes #1001

### DIFF
--- a/src/components/auth/ProtectedRoute.tsx
+++ b/src/components/auth/ProtectedRoute.tsx
@@ -13,6 +13,7 @@ const ProtectedRoute = ({ children }: ProtectedRouteProps) => {
   const [session, setSession] = useState<Session | null>(null);
   const [loading, setLoading] = useState(true);
   const [authError, setAuthError] = useState<string | null>(null);
+  const [needsMfaChallenge, setNeedsMfaChallenge] = useState(false);
 
   useEffect(() => {
     let isMounted = true;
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #1012, opened 2026-08-06
fix: guard profile field decryption so one bad field does not break the page (#1002)

### LINKED ISSUE #1002: [BUG] Profile page fails entirely with Connection Error when a single profile field cannot be decrypted
﻿The Profile load runs all field decryptions inside one try/catch. `decryptText` throws `Invalid encrypted text format` for any value that is not `ivHex:cipherHex`, and `decryptProfileField`/`decryptProfileArray` have no per-field guard or plaintext passthrough. If a single stored field is plaintext, malformed, or encrypted under a different key (e.g. after the salt/key-rotation problems, or legacy data), the whole page load throws and the user only sees "Connection Error - Failed to load your profile" ? the entire page is unusable.

Location:
- `src/pages/Profile/index.tsx:58-91` ? all `decryptProfileField`/`decryptProfileArray` calls inside one try/catch
- `src/lib/encryption.ts:245-249` ? `decryptText` throws on any value that does not split into exactly 2 parts
- `src/lib/encryption.ts:545-560` ? `decryptProfileField` / `decryptProfileArray` call `decryptText` directly with no fallback

Suggested fix: guard each field (return "" / [] for missing or non-encrypted values) and surface per-field errors instead of failing the whole load.

ionfwsrijan:
## Problem

`decryptProfileField` and `decryptProfileArray` throw when given a missing, non-encrypted, or malformed value (e.g. a user who never filled in a field, or an old row stored before encryption was added). A single bad field aborts the entire profile load, so the whole Profile page breaks instead of showing the rest of the data.

## Fix

- `decryptProfileField` returns `""` and `decryptProfileArray` returns `[]` for missing/non-encrypted/malformed values instead of throwing.
- Adds an `isEncryptedProfileValue` guard that only attempts decryption for values in the expected `enc:` format.
- Wraps each field's decryption in a per-field `try/catch` that logs a warning, so one bad field no longer aborts the rest of the profile.

## Files changed

- `src/lib/encryption.ts`

## Testing

- Manual QA: open the Profile page for a user with a missing/malformed encrypted field and confirm the page still loads with the remaining fields populated.

Closes #1002

### DIFF
--- a/src/lib/encryption.ts
+++ b/src/lib/encryption.ts
@@ -547,7 +547,13 @@ export async function decryptProfileField(
   key: CryptoKey
 ): Promise<string> {
   if (!value) return "";
-  return await decryptText(value, key);
+  if (!isEncryptedProfileValue(value)) return "";
+  try {
+    return await decryptText(value, key);
+  } catch (err) {
+    console.warn("Failed to decrypt profile field, returning empty value:", err);
+    return "";
+  }
 }
 
 export async function decryptProfileArray(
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #1013, opened 2026-08-06
fix: restore AAL2 session when delete-account verification fails (#1003)

### LINKED ISSUE #1003: [BUG] Delete-account verification replaces the current session with an unverified AAL1 session (MFA downgrade)
﻿To confirm account deletion, `handleDeleteAccount` calls `signInWithPassword`, which replaces the user's current (AAL2) session with a fresh password-only AAL1 session. If the subsequent `delete-user-account` edge function call fails, the user remains signed in under the weaker AAL1 session, and nothing in the app re-challenges 2FA (`ProtectedRoute` only checks session existence).

Location:
- `src/pages/User/Settings.tsx:164-178` ? `signInWithPassword({ email, password: deletePassword })`
- `src/pages/User/Settings.tsx:183-184` ? on edge-function failure the user stays signed in
- `src/components/auth/ProtectedRoute.tsx:20-33` ? no AAL2 check anywhere else in the app

Impact: for MFA users, a failed deletion attempt silently downgrades the session to password-only.

Suggested fix: use `supabase.auth.reauthenticate()` (with an MFA challenge where required) instead of replacing the session with a fresh unverified sign-in.

ionfwsrijan:
## Problem

The delete-account flow calls `signInWithPassword` to re-verify the user's password. If the verification fails or the `delete-user-account` edge function returns an error, the user is left on an AAL1 session — the original AAL2 (2FA-verified) session is lost, so the user silently loses MFA protection on the current device.

## Fix

- Captures the current session (including its AAL2 assurance level) before re-verifying the password.
- If password verification fails or the `delete-user-account` call fails, restores the captured session via `supabase.auth.setSession` so the user keeps their 2FA-verified session.

## Files changed

- `src/pages/User/Settings.tsx`

## Testing

- Manual QA: with MFA enabled, attempt to delete the account and enter a wrong password; confirm the current session remains AAL2 and the user is not logged out or downgraded.

Closes #1003

### DIFF
--- a/src/pages/User/Settings.tsx
+++ b/src/pages/User/Settings.tsx
@@ -162,14 +162,26 @@ const Settings = () => {
         return;
       }
 
-      // Verify password
+      // Verify password. signInWithPassword replaces the active session with a
+      // fresh password-only AAL1 session, so capture the current (AAL2)
+      // session first and restore it whenever the deletion does not go through
+      // — otherwise MFA users would be left signed in under the weaker session.
       const user = (await supabase.auth.getUser()).data.user;
+      const { data: sessionData } = await supabase.auth.getSession();
+      const currentSession = sessionData.session;
+
       const { error: signInError } = await supabase.auth.signInWithPassword({
         email: user?.email || "",
         password: deletePassword,
       });
 
       if (signInError) {
+        if (currentSession) {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #1014, opened 2026-08-06
fix: use local dates for check-in challenge and streaks (#1004)

### LINKED ISSUE #1004: [BUG] Challenge daily check-in uses UTC date so the check-in day differs from the user local calendar day
﻿The daily challenge check-in uses the UTC date (`now.toISOString()`) to decide whether the user has already checked in "today", so the check-in window is shifted relative to the user's local calendar day.

Location:
- `src/hooks/useGamification.ts:142` ? `const todayDateStr = now.toISOString().split("T")[0];`
- `src/hooks/useGamification.ts:143` ? `todayStartISO = \`${todayDateStr}T00:00:00.000Z\``
- `src/hooks/useGamification.ts:145-147` ? `alreadyCheckedInToday = lastCheckIn.toISOString() >= todayStartISO`
- `src/hooks/useGamification.ts:153-163` ? streak day-count also computed from UTC date strings

Impact: for users east of UTC the local calendar day rolls over while UTC is still on the previous day (e.g. UTC+5:30 users cannot check in between local 00:00-05:30 because the UTC date is still yesterday, so `alreadyCheckedInToday` returns true for a brand-new local day); for users west of UTC the same UTC day spans two local days, allowing extra/missed check-ins. Streak counting is therefore timezone-dependent.

Suggested fix: compute the check-in day boundary from the user's local time (e.g. derive `todayStartISO` from the local date) or store the intended local date explicitly.

ionfwsrijan:
## Problem

The daily check-in challenge and streak logic compare dates in UTC. For users in timezones ahead of UTC this can mark a check-in as "tomorrow", break an existing streak, or allow two check-ins on the same local day — the check-in experience is wrong for most timezones.

## Fix

- Adds a `toLocalDateStr` helper that formats a timestamp into the user's local date.
- Computes `todayStartISO` from the start of the local day instead of UTC.
- Streak calculations now compare local dates, so a check-in always counts toward the correct local day.

## Files changed

- `src/hooks/useGamification.ts`

## Testing

- Manual QA: set the device timezone ahead of UTC and confirm a check-in counts for today's local date and the streak is maintained correctly.

Closes #1004

### DIFF
--- a/src/hooks/useGamification.ts
+++ b/src/hooks/useGamification.ts
@@ -139,8 +139,16 @@ export function useCheckInChallenge() {
       if (fetchErr) throw fetchErr;
 
       const now = new Date();
-      const todayDateStr = now.toISOString().split("T")[0]; 
-      const todayStartISO = `${todayDateStr}T00:00:00.000Z`;
+      const toLocalDateStr = (date: Date): string =>
+        `${date.getFullYear()}-${String(date.getMonth() + 1).padStart(2, "0")}-${String(
+          date.getDate()
+        ).padStart(2, "0")}`;
+
+      // Compute the check-in day boundary from the user's local calendar day
+      // instead of the UTC date, so the check-in window matches the day the
+      // user sees on their device.
+      const todayDateStr = toLocalDateStr(now);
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #1015, opened 2026-08-06
fix: delete-all-history uses authenticated user id and handles zero rows (#1005)

### LINKED ISSUE #1005: [BUG] Delete All history no-ops the server delete and shows success when the visible list is empty
﻿`deleteAllEntries` targets `history[0]?.user_id || ""`. The "Delete All" button is always rendered (not gated on list length), so when the visible (filtered/search) list is empty ? e.g. every record is filtered out by a severity filter ? `user_id` becomes `""`, the server delete matches zero rows, the local Dexie cache is still cleared, and a success toast is shown. The records reappear on the next fetch because they were never deleted.

Location:
- `src/pages/History/index.tsx:324-343` ? delete-all handler
  - line 330: `.eq("user_id", history[0]?.user_id || "")`
  - line 334: `await db.symptomHistory.clear()` ? unconditional local wipe
- `src/pages/History/index.tsx:481-487` ? Delete All button always visible in the header

Suggested fix: use the authenticated user id from `supabase.auth.getUser()` instead of the first visible record, and do not claim success for a zero-row delete.

ionfwsrijan:
## Problem

`deleteAllEntries` deletes by `history[0]?.user_id || ""`. After the user clears their history locally (empty list), `user_id` resolves to `""`, so the server delete targets `user_id = ""`, affects zero rows, does not raise an error, and the UI shows a false-success toast and wipes state that was never actually deleted.

## Fix

- Resolves the authenticated user via `supabase.auth.getUser()` and deletes by that real `user_id`.
- Checks the deleted-row count from `.select("id")`; a zero-row delete shows an informational toast (instead of a success toast) and returns early, while a successful multi-row delete keeps the existing success behavior.

## Files changed

- `src/pages/History/index.tsx`

## Testing

- Manual QA: with history present, delete all and confirm rows are removed and the success toast shows; with an empty history, confirm the info toast appears instead of a false success.

Closes #1005

### DIFF
--- a/src/pages/History/index.tsx
+++ b/src/pages/History/index.tsx
@@ -19,7 +19,7 @@ import jsPDF from "jspdf";
 import autoTable from "jspdf-autotable";
 import { useNavigate } from "react-router-dom";
 import { useToast } from "@/hooks/use-toast";
-import { showSuccess, showError } from "@/lib/toast-helpers";
+import { showSuccess, showError, showInfo } from "@/lib/toast-helpers";
 import { db, syncOfflineData, encryptSymptom, decryptSymptom } from "@/lib/offline-db";
 import { whenKeysReady, generateSearchTokens } from "@/lib/encryption";
 import { getCachedData, invalidateCache } from "@/lib/cached-queries";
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

Issue #1030, opened 2026-08-06
[BUG] <Brief Description>A box appears while writing anything in the ai assisstant colour varies according to the theme the box needs to be removed


ritesh123malik:
## 🔄 Steps to Reproduce  
1. **Go to:** "..."  
2. **Click on:** "..."  
3. **Expected Outcome:** "..."  
4. **Actual Outcome:** "..."  

## 📌 System Information  
- **Operating System:** (Windows/macOS/Linux)  
- **Browser/Device:** (Chrome, Firefox, Safari, etc.)

PR #1038, opened 2026-08-06
fix(assistant): remove focus ring box from symptoms text input #1030

ritesh123malik:
Resolves issue #1030. Overrides the global focus-visible box-shadow ring on the native symptoms textarea in the AI Health Assistant page to prevent the inner outline from appearing when typing.


ritesh123malik:
@
Screenshot 2026-08-06 at 8 13 00 PM
after

PR #1066, opened 2026-08-08
fix: decrypt errors with DecryptionError class

### LINKED ISSUE #910: fix : normalize decrypt errors with DecryptionError class

The `decryptText` function in `src/lib/encryption.ts` throws raw `DOMException` (or `TypeError`) from `crypto.subtle.decrypt` when decryption fails (wrong key, corrupted data, tampered ciphertext). Callers who want to handle decryption failures gracefully must know about browser-specific error types, making the API brittle and error-handling inconsistent across the codebase.

- Create a `DecryptionError` class that extends `Error` with `name = "DecryptionError"`
- Wrap the `crypto.subtle.decrypt` call in a `try/catch`
- Throw `DecryptionError` (not raw DOMException) on any decryption failure
- Export `DecryptionError` from the module so callers can `instanceof` check it
- Keep the existing `Invalid encrypted text format` check as a `DecryptionError` as well

- Consistent, typed error surface for all decryption failures
- Callers can catch `DecryptionError` without browser-specific instanceof checks
- Makes the encryption module easier to test and reason about

Note: Please assign this issue to the `tmdeveloper007` account.

Aditya8369:
Summary of Changes
Created & Exported DecryptionError: Added DecryptionError extending Error with name = "DecryptionError".
Normalized decryptText Errors: Wrapped the crypto.subtle.decrypt call in a try/catch block to catch underlying Web Crypto failures (DOMException, TypeError, etc.) and throw a normalized DecryptionError with cause set to the underlying error.
Format Check: Updated the invalid format check (parts.length !== 2) in decryptText to throw DecryptionError("Invalid encrypted text format").
Unit Tests: Added test cases in src/lib/encryption.test.ts
 verifying that format errors and decryption failures (e.g. wrong key) throw an instanceof DecryptionError with name === "DecryptionError".

closes #910

### DIFF
--- a/src/lib/encryption.test.ts
+++ b/src/lib/encryption.test.ts
@@ -6,6 +6,7 @@ import {
   whenKeysReady,
   encryptText,
   decryptText,
+  DecryptionError,
   deriveKeyFromToken,
   deriveSearchKeyFromToken,
   rotateKeysToNewPassword,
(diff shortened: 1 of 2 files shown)


Aditya8369:
@mohdmaazgani please check this

Issue #1157, opened 2026-08-09
[BUG] <Brief Description>currently on exporting the data on google sheets the data is not correctly getting exported


ritesh123malik:
## 🔄 Steps to Reproduce  
1. **Go to:** "..."  
2. **Click on:** "..."  
3. **Expected Outcome:** "..."  
4. **Actual Outcome:** "..."  

## 📌 System Information  
- **Operating System:** (Windows/macOS/Linux)  
- **Browser/Device:** (Chrome, Firefox, Safari, etc.)

mohdmaazgani:
@ritesh123malik I guess this issue is still not completed. Please complete this as soon as possible.

SinghAnsh07:
I would like to work on this. Can you please assign it to me under GSSoC'26 contributor!?

PR #1158, opened 2026-08-09
fix: resolve hidden date columns and stray asterisks in CSV export (#1157)

ritesh123malik:
This PR resolves the two bugs reported in #1157:

1. **Hidden date columns**: Excel often displays  when localized date formats exceed the column width. We standardise the formatting to use the short `en-US` locale (`M/D/YYYY` format) and prepend a UTF-8 Byte Order Mark (`\ufeff`) to the CSV blob, resolving decoding and display quirks in Excel.
2. **Stray asterisks**: We now strip any markdown bold/italic formatting tags (`**` or `*`) from symptoms, possible causes, and recommendations before exporting.

### DIFF
--- a/src/pages/History/index.tsx
+++ b/src/pages/History/index.tsx
@@ -350,30 +350,57 @@ const History = () => {
 
   const exportCSV = () => {
     const headers = [
-  "Date",
-  "Symptoms",
-  "Severity",
-  "Risk Score",
-  "Possible Causes",
-  "Recommendations",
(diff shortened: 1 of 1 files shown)


mohdmaazgani:
@ritesh123malik The astreick problem has been solved, but not the date one of ######...please carefully solve that 

image

mohdmaazgani:
and make sure to change the format to DD/MM/YYYY

PR #1174, opened 2026-08-11
security(1056): remove userId-derived fallback key; require password …

### LINKED ISSUE #1056: [SECURITY] Fallback encryption key is derived from the non-secret `userId` when no master seed is persisted
## 🐛 Bug Description  

In `src/lib/encryption.ts` → `handleSessionChange`, when no persisted master seed exists for the session:
```ts
const storedSeed = localStorage.getItem(SEED_KEY_PREFIX + userId);
const masterSeed = storedSeed || userId;   // ← non-secret fallback
const newKey = await deriveKeyFromToken(masterSeed, userId);
```

The AES-GCM key is then derived from  userId  (public) plus the per-user PBKDF2 salt, which is also stored in the user's  profiles.encryption_salt  row and auth user metadata. Anyone with read access to those values (or a misconfigured RLS policy / service-role context) can re-derive the key and decrypt health records that were written under the fallback key. The code itself flags this: "derives a weaker, non-secret-based key, since userId is not a secret".

## Who is affected
- Accounts created before the master-seed mechanism existed (their historical records are keyed by  userId ).
- Sessions where the local seed was cleared (e.g. sign-out on another tab) while the session remains active.

## 🔄 Steps to Reproduce  
1. **Go to:** "..."  
2. **Click on:** "..."  
3. **Expected Outcome:** "..."  
4. **Actual Outcome:** "..."  

## ✅ Expected Behavior  

Keys must never be derived from public-only material; without the seed, the app should require the user to re-enter their password instead of silently degrading.

## 🛠️ Possible Fix (if known)  

Remove the  userId  fallback and gate key derivation on the persisted seed (re-prompt for password when absent), or escrow-encrypt the seed under a device-local non-extractable key before it can be used as a fallback.

## 📌 System Information  

gssoc26

vib3withsimran:
…dexedDB

Fixes a security issue where the AES-GCM encryption key was derived from the **public `userId`** whenever no persisted master seed was available: `const masterSeed = storedSeed || userId` in `handleSessionChange()`. Because the per-user PBKDF2 salt is also public (synced to `profiles.encryption_salt` and auth user metadata), anyone with read access to those values could re-derive the key and decrypt health records written under the fallback key.
 1. **`src/lib/encryption.ts`**
   - Removed the `userId` fallback in `handleSessionChange()`. When a session exists but no master seed is persisted, the keys are now **locked** (`setKeys(null, null)`) instead of being derived from public material — satisfying "keys must never be derived from public-only material".
   - New `subscribeEncryptionLock()` API so the UI can react to the locked state.
   - New `unlockEncryptionWithPassword(password)` flow: validates the password via Supabase re-auth (same pattern as the Settings page — a wrong password must never silently derive a wrong seed), persists the seed via `setupKeysFromPassword`, and activates the seed-derived keys.
   - **Legacy migration**: for accounts created before the master-seed mechanism (whose records are encrypted under the userId-derived key), the unlock re-derives the legacy key, detects such records via a registered probe, and rotates them onto the new seed-derived key using the existing `triggerKeyRotation` re-encryption machinery — no data loss.
   - Race guards: `unlockInProgress` (the re-auth call fires auth events that must not clobber the in-flight unlock), freshly-set keys on normal sign-in, and user-switch clearing.
 
2. **`src/lib/offline-db.ts`**
   - Local `onTokenRefresh` re-encryption changed from "clear all tables on failure" to **per-record skip** — a key rotation can never destroy local data.
   - Registers the legacy-key probe (local IndexedDB + server-side rows; skips plaintext records to avoid false positives).
 
3. **New `src/components/auth/EncryptionUnlockDialog.tsx`** — a non-dismissible dialog shown app-wide whenever keys are locked. It asks for the account password, shows validation errors, and offers a **"Forgot your password? Sign out"** escape hatch so no user is ever trapped (sign-out clears the lock and returns to the sign-in page).
 
4. **`src/App.tsx`** — renders the unlock dialog; **`src/components/ui/dialog.tsx`** — added an optional `hideCloseButton` prop.
 
5. **Tests** (`src/lib/encryption.test.ts`) — 5 new tests covering: keys lock (not derived) when a session has no seed, successful unlock + seed persistence, incorrect-password rejection, legacy-data migration, and the no-session error path.
 
---
 
### **4. Impact / Verification**
 
**Behavior**
- Normal users are unaffected: sign-in always persists the seed, so keys derive exactly as before.
- Users who lose the seed (cleared storage, sign-out on another tab) now see an unlock dialog instead of silently using an insecure key. Re-entering the password re-derives the same deterministic seed, so existing data remains readable.
- Legacy (pre-seed) accounts are migrated to the new key on unlock, preserving access to historical records.
- The unlock migration derives the legacy key **only transiently in memory** to move already-exposed records onto the new key — no new exposure is introduced.
---

### **1. Type of Change**

---

### **2. Related Issue**
- Fixes #1056 

---

- [x] **Local Build**: The application builds successfully locally (`npm run build`).
- [x] **Lint/Format Checks**: Local checks passed (`npm run lint` / `npm run format:check`).

---

| Before | After |
| :--- | :--- |

---


mohdmaazgani:
@vib3withsimran Please resolve the merge conflicts.

